PR #4, opened 2019-07-29
D8 nid 250

DuttonMa:
Permissions changes and other tweaks to workflow views.


config/sync/user.role.gp_author_user.yml
```diff
@@ -8,8 +8,11 @@ weight: 2
 is_admin: null
 permissions:
   - 'access toolbar'
+  - 'add gp entities'
   - 'add media from remote sources'
+  - 'administer gp entities'
```
johangant: Wouldn't 'administer gp entities' be better restricted to site admins?

DuttonMa: Ah yes, you're quite correct. I was messing around with permissions in an effort to give users the ability to access the 'Content / Add GP' admin menu but wasn't able to get it to work.
In the end I created a Jira ticket for all of the permission changes that we need.
I'll remove this permission from the gp_author.

PR #7, opened 2019-08-02
Add contrib module dependencies and things to make phpunit usage easier

johangant:
Oddly, search_api installation with composer doesn't list drupal/facets as a dependency. While functional, if you ask PHPUnit to list groups the former fails on missing Traits provided by the latter making testing impossible. Adding the module as a dependency in composer fixes this.

Ditto for metatag and schema_metatag modules.

drupal-check and phpunit-finder added to require-dev as they're useful when you want to avoid global installs on CI services and a bit more intel about the test suites to help to split them up across multiple containers.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -74,6 +74,7 @@
         "drupal/console": "^1.0.2",
         "drupal/core": "^8.7.0",
         "drupal/entity_embed": "1.0",
+        "drupal/facets": "^1.4",
         "drupal/field_group": "^1.0",
         "drupal/flag": "4.x-dev",
         "drupal/geolocation": "^1.11",
@@ -88,6 +89,7 @@
         "drupal/paragraphs_inline_entity_form": "^1.0@beta",
         "drupal/pathauto": "^1.4",
         "drupal/redirect": "^1.3",
+        "drupal/schema_metatag": "^1.4",
         "drupal/search_api": "^1.14",
         "drupal/search_api_autocomplete": "^1.2",
         "drupal/search_api_solr": "^3.2",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -116,6 +118,8 @@
         "drupal/migrate_upgrade": "^3.0",
         "drupal/restui": "^1.17",
         "drupal/stage_file_proxy": "^1.0@beta",
+        "mglaman/drupal-check": "^1.0",
```
omahm: Drupal check is installed as part of the lando build process. I originally had it as a composer dependency but had to remove it and pull from the repo as the composer version was broken. Have you verified that the composer version works?

johangant: It seems to. Or at least, it's struggling with the exact same issues around autoloading some of the classes (think I fixed late last week).

It also works with the global installer mind you... if memory serves I believe I hit problems with autoloading test classes and wondered whether this was due to having a global install of the tool. Moving it into composer felt like a natural first step given that drush and other tools are used in this way to avoid any global config issues.

What I can do today is grab the PRs from upstream, get them to apply and see if this works (or doesn't). While it's nice to have Lando + CI use the same kind of approach, it's probably not that big a deal in practice as Lando fetches the latest release and composer would do something similar... and there's no major issue with global tools on the CI container as its spun up/torn down pretty fast with no cross-over with other apps.

omahm: Sounds like a plan, I'd check out drupal-check via git and composer and see how current the composer version is as that's what caused me a load of grief a few months back. It's still  in its infancy so a lot of updates and patches are absent from the composer release versions.

johangant:
Looks like it's working fine to me with identical output from global install plus composer install. I guess we can always pull it out of composer + use a global install if something breaks but as far as I can tell it looks stable with these with the PR to add fixes for the custom code repo.

PR #9, opened 2019-08-07
Add CircleCI pipeline and docs

johangant:
This PR provides a CI pipeline that:

- Checks the integrity of the build
- Checks compliance with Drupal coding standards
- Checks for deprecated code
- Runs unit and kernel tests on a vanilla D8 site with our site config applied to it.
- Runs UI tests via nightwatch.js

UI tests are presently stubbed out as they need a little more work from services to provide things like safe DB dump files and a decision about whether to test migrated content against the presently live site.


johangant:
Obvs I will be squashing commits in this PR! Debugging locally wasn't always trivial and making small incremental changes to run it on the more efficient CircleCI platform can get pretty tedious

phpcs.sh
```diff
@@ -23,6 +23,8 @@ DRUPAL_EXCLUDED_SNIFFS=(
 # IGNORE="$IGNORE,${DRUPAL_DEPLOY_PATH}/web/themes/custom/nidirect/node_modules"
 
 echo "Running coding standard checks in ${PHPCS_CHECK_DIR}"
+echo "PHPCS_PATH: ${PHPCS_PATH}"
```
johangant: Eeep. Shouldn't be there, let me pull that out now

johangant:
I should add that there are some refinements that could still be made. Namely:

- YAML references/anchors to make things a little more DRY. They were kinda fussy and I opted for clarity over brevity.
- Environment variables/interpolation in commands/steps is pretty fussy. I found the $CIRCLE_WORKING_DIRECTORY didn't always work... so opted for absolute paths where I needed them. Again, just for clarity while getting things working.

.circleci/config.yml
```diff
@@ -0,0 +1,274 @@
...
+              cp /home/circleci/project/.circleci/drupal.settings.php sites/default/settings.php
+              # Copy PHPUnit config into core folder.
+              cp /home/circleci/project/.circleci/phpunit.circleci.xml core/
+              # Install Drupal (or restore from DB dump at this point).
+              ../vendor/bin/drush site-install -y standard --site-name="NI Direct"
+              # Enable migrate_nidirect_utils.
+              ../vendor/bin/drush en -y migrate_nidirect_utils
+              # No migrate db, so fudge the drupal-console commands into drush equivalents.
+              ../vendor/bin/drush cset -y "system.site" uuid "${UUID}"
+              ../vendor/bin/drush ev '\Drupal::entityManager()->getStorage("shortcut_set")->load("default")->delete();'
```
omahm: This will blow up. Drupal will complain if the default shortcut set is missing. You need to load each link belonging to that set and delete each of those.
See: https://github.com/dof-dss/nidirect-d8-mig-mods/blob/development/migrate_nidirect_utils/src/Command/NidirectMigratePreCommand.php#L73

johangant: You can probably tell from the comment I tried to use drupal console to run this, but because I was missing the migrate DB config it didn't want to play any further. I quite like the idea of re-using the code for that command, so I'm tempted to bash in some placeholder values for the config it needs and see how it behaves. That said, those two commands where what I've run before we had the drupal console commands and didn't seem to cause any trouble... but perhaps I've missed a step after that and totally not seen a bigger problem :)

omahm: I'd originally deleted the default shortcut set thinking it would be ok but once you start trying to use the site it'll throw up a routing error.

johangant: Okay, so I can't use drupal console directly as I'm missing a bunch of stuff we'd define in Lando. Even if I fake it with placeholder values the command I want to use starts wanting to fudge source db URL aliases and all the other helpful things that are going in when it comes to local dev work.

What I've done is set up a small script I can run in isolation with drush that borrows the bits of code from that command that we need. So the outcome is the same, it safely removes shortcut links, at the tiny expense of not being as DRY and using the command itself via drupal console. A bit irritating, but I can live with that for now. I guess we can create a CI helper command in future if that's more appropriate but I don't see it being a critical issue.

johangant:
Config uuid and shortcut handling now resolved. I think that covers all issues so far, so will get this merged in

PR #10, opened 2019-08-09
D8NID-273 Config split

johangant:
This PR aims to:

- Set out config split configurations for environments:
  - Local development
  - Development 
  - Pre-production
  - Production (`sync`)
- For local development:
  - Keep an initial blacklist (totally separate) for modules like devel.
  - Keep a graylist (merged with main config in `sync`) for enabled modules.
- Improve the documentation to show how config_split is configured, and why.
- Illustrate how to use it on a day to day basis, along with any caveats.


johangant:
> Are the migration config files needed? The config will be imported when the migration modules are enabled.

You're probably right here. That said, once you have them enabled and the config is active... any config export will always export so would need developers to avoid re-committing them which isn't a particularly safe or robust approach.

Perhaps config_ignore could help but ideally, if the config filters can be tweaked to avoid dumping those migration and migration group config files then that might be more useful.

omahm:
> Perhaps config_ignore could help but ideally, if the config filters can be tweaked to avoid dumping those migration and migration group config files then that might be more useful.

Yeah, I'd prefer to go down that route of using config ignore.

DuttonMa:
I'm inclined to agree, so far we have had it all working nicely when we create an empty D8, import config and then enable migration modules. We have thus managed to avoid having the migration configs in config/sync and that does feel like the best way to carry on.
We have so far managed to avoid committing the migration configs so we should be able to carry on that way.

johangant:
Cool, sounds like it's going in the right direction. Have just pushed an update that replaces 'local_development' with just 'local'

Config ignore will now never import migrate_plus or migrate_drupal config from the filesystem - if you're debugging migrations in future you might want to turn that off, or have a config_ignore config setting in 'local' to allow you to do that.

I haven't yet found a way to stop the migrate config files from exporting once the modules are enabled. I don't think this is a huge problem - if it is I'd suggest we don't include them in a given config split unless they're actually needed

PR #13, opened 2019-08-14
Enable nics theme for local development

DuttonMa:


config/local/system.theme.yml
```diff
@@ -0,0 +1,4 @@
+admin: seven
```
johangant: I'm thinking this will change to be adminimal_theme in due course, but fine with me to leave it at seven for now

PR #20, opened 2019-08-24
Enable social sharing tools module

johangant:
NB: expecting this to be revisited/rejigged a little once the config split settings merge into development branch


johangant:
> Think we need to remove core.extension.yml from the /local directory

@DuttonMa now done 👍

johangant:
Marking as prospective future enhancement and closing.

PR #24, opened 2019-08-28
Add lock file to allow composer install to fetch new dependencies

johangant:
I noticed that https://github.com/dof-dss/nidirect-drupal/pull/23/files?file-filters%5B%5D= was merged, but `composer install` didn't fetch the new project for me. I think it needs the lock file to go with it as that's what composer install reads from to fetch dependencies.


johangant:
@DuttonMa has beaten me to it, closing this

PR #28, opened 2019-08-29
further path changes

DuttonMa:
- Change remaining 'nidirect_workflow' paths to 'nicsdru_workflow'
- Workflow permissions fix


omahm:
Looks good but we should omit the 'mas_rss' perms as we no longer use Feeds to pull these in.

PR #33, opened 2019-09-03
D8 nid 295

omahm:
Not an exhaustive list of restrictions as we still need to define layouts and blocks etc.

### DIFF
--- a/config/sync/core.extension.yml
+++ b/config/sync/core.extension.yml
@@ -39,6 +39,7 @@ module:
   filter: 0
   flag: 0
   geolocation: 0
+  handy_cache_tags: 0
   help: 0
   hms_field: 0
   image: 0
@@ -98,6 +99,7 @@ module:
   twig_extensions: 0
   ultimate_cron: 0
   user: 0
+  views_custom_cache_tag: 0
   webform: 0
   workflows: 0
   menu_link_content: 1
--- a/config/sync/core.extension.yml
+++ b/config/sync/core.extension.yml
@@ -37,13 +37,16 @@ module:
   filter: 0
   flag: 0
   geolocation: 0
-  handy_cache_tags: 0
   help: 0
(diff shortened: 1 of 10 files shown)


config/sync/core.extension.yml
```diff
@@ -39,13 +39,16 @@ module:
   filter: 0
   flag: 0
   geolocation: 0
-  handy_cache_tags: 0
```
johangant: I think we want to keep this module enabled: it's handy to provide more granular cache tags for views listings

config/sync/core.extension.yml
```diff
@@ -95,7 +98,6 @@ module:
   twig_extensions: 0
   ultimate_cron: 0
   user: 0
-  views_custom_cache_tag: 0
```
johangant: As above: I think we want to keep this module enabled: it's handy to provide more granular cache tags for views listings.

config/sync/ultimate_cron.job.layout_builder_cron.yml
```diff
@@ -0,0 +1,17 @@
+uuid: 7c618462-dfc2-4ba0-8eb0-608c1c5273ff
```
johangant: Does layout builder need cron? I'm guessing this is auto-exported but wondering what things you'd want to schedule with it?

config/sync/core.extension.yml
omahm: Good catch, recently added modules. This will be one thing to watch as the pace picks up.

config/sync/ultimate_cron.job.layout_builder_cron.yml
omahm: It adds an entry into ultimate cron. Complete guess but it's possibly in there to do routine cleanup of displays/layouts etc.

PR #37, opened 2019-09-05
Update coder module for latest code sniffs.

johangant:
- Add DrupalPractice.Objects.StrictSchemaDisabled to ignore list

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -9419,16 +9419,16 @@
         },
         {
             "name": "symfony/yaml",
-            "version": "v3.4.30",
+            "version": "v3.4.31",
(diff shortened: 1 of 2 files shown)


johangant:
Yay for failing tests on things to fix failing tests :)

I've temporarily unfollowed the project on CircleCI until this PR is in as this will enable the wider workflow to follow. Presently it'll either fail at static analysis or possibly even SSH keys if they need a look again

johangant:
Yeah I kind of need this to stop things from being flagged as broken. If it needs a tweak again, I'll fix next week 👍

PR #48, opened 2019-09-20
Add config for search autocomplete and block

johangant:
There's a corresponding PR in the site-modules repo that provides a block for the views embed display to sit inside. Using views block display means you can't get the form to submit to a different page as the AJAX behaviour reloads the same page on submit.

### DIFF
--- a/config/sync/block.block.sitesearchblock.yml
+++ b/config/sync/block.block.sitesearchblock.yml
@@ -1,4 +1,4 @@
-uuid: bcd31f12-3b6f-4852-be60-03e09b9a8818
+uuid: c86287b8-2f26-4e18-a9a9-fb026899ea98
 langcode: en
 status: true
 dependencies:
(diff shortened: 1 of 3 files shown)


config/sync/block.block.sitesearchblock.yml
```diff
@@ -0,0 +1,26 @@
+uuid: bcd31f12-3b6f-4852-be60-03e09b9a8818
+langcode: en
+status: true
+dependencies:
+  module:
+    - nidirect_common
+    - system
+  theme:
+    - bartik
```
omahm: Should be nidirect, or can this not be tested with in its current state?

johangant: Not yet, it doesn't give a very helpful presentation. Once it's fleshed out more I'm happy to move this into the right place.

@David-Corbett if you're already looking at page furniture for the nidirect theme, once this and its corresponding PR are merged you can use the `site_search_block` to add to any locations it might be needed.

johangant: Now done + pushed

PR #52, opened 2019-09-23
D8 nid 223

omahm:


config/sync/views.view.featured.yml
```diff
@@ -0,0 +1,226 @@
...
+          table: node_field_data
+          field: status
+          plugin_id: boolean
+          entity_type: node
+          entity_field: status
+          id: status
+          expose:
+            operator: ''
+          group: 1
+        promote:
```
johangant: There's a flag in D7 for indicating featured content. Might be worth clarifying if that's used or whether we can get away with using the default promoted checkbox on nodes

johangant: Oh hang on, I see the flag defined too. How is content indicated as being featured? Flag + promoted or can it be one or the other too?

omahm:
> Makes sense to me. I wasn't too sure how featured content is defined between the flag and the promoted checkbox on nodes... are there clear differences between the two?

Yeah, it's a bit of an oddity and I was a bit wary of changing the current editorial workflow. So it'll show stuff that's flagged as featured and promoted. Looks to me like the featured flag is then used in 'Featured' areas within in sub-sections. I think we could possibly get away with 'Promoted' only but then they only want certain types promoted so I'd need a filter to restrict content types. The current 'featured' flag acts as a filter and ideally we'd have a play around with the UI and test it with CET.

omahm:
After discussion with the rest of the team, we are going to take a different approach.

PR #53, opened 2019-09-23
Add perms filter module

johangant:
Long + wide permissions grids make for a hard time picking out specific things. This module really helps cut through the clutter, I'd highly recommend we incorporate it into the build as standard (benefits site admins + devs) or at the very, very least as a development tool.


composer.json
```diff
@@ -94,6 +94,7 @@
         "drupal/paragraphs": "^1.8",
         "drupal/paragraphs_inline_entity_form": "^1.0@beta",
         "drupal/pathauto": "^1.4",
+        "drupal/permissions_filter": "^1.1",
```
omahm: I think this should go in as a dev requirement, we're not gonna need that on prod?

johangant: _makes thinking noises_.... yeah, I guess if perms are controlled in config there's less scope for admins in comparison to a classic D7 site.

Dev requirement it is then 👍

johangant: All done now with tests passing

PR #57, opened 2019-09-30
D8NID-258 themes listing

johangant:
See https://github.com/dof-dss/nidirect-site-modules/pull/34 for overview of how this fits in with the custom module.


johangant:
NB: Kernel/Unit tests won't presently pass because https://github.com/dof-dss/nidirect-site-modules/pull/34 hasn't been merged so the custom module doesn't exist until it has.

PR #65, opened 2019-10-04
Base theming - set main blocks in regions

David-Corbett:
set some of the main system blocks in nidirect theme to make help the site work / look like it should


config/sync/migrate_plus.migration_group.default.yml
```diff
@@ -1,4 +1,4 @@
-uuid: 7eb6f4eb-6108-4699-939f-063763ad2ae5
```
omahm: Is this mean to be in here?

David-Corbett: no, have fixed this

PR #67, opened 2019-10-09
uninstall all migrate modules

DuttonMa:
As discussed yesterday, uninstall all migrate modules from config and add migrate config files to .gitignore.
Related to https://github.com/dof-dss/nidirect-d8-mig-mods/pull/11


config/sync/core.extension.yml
```diff
@@ -63,26 +63,6 @@ module:
   metatag_verification: 0
   migrate: 0
   migrate_drupal: 0
```
johangant: is `migrate_drupal` a dependency of something like money advice service? Not a problem if it is, but I would imagine this exclusively migrate related so I'd consider removing from here and add as a dependency of the global migration?

PR #69, opened 2019-10-11
D8NID-335 ~ Update origins_modules to 0.1.2 and apply patch

omahm:


omahm:
> Looks good, but why a patch rather than rolling a new release ?

It's a tiny alteration and to me, doesn't warrant a full release. It'll be fixed in the next and then we can pull the patch from composer.

PR #75, opened 2019-10-18
D8NID-346 Add scheduler module and initial config

johangant:
Applied to all node content types with publish + unpublish options set.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -96,6 +96,7 @@
         "drupal/paragraphs_inline_entity_form": "^1.0@beta",
         "drupal/pathauto": "^1.4",
         "drupal/redirect": "^1.3",
+        "drupal/scheduler": "^1.1",
         "drupal/schema_metatag": "^1.4",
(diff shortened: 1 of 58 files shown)


johangant:
> I don't understand why we have all of these 'promote to front page' Yaml files in this change, are they really necessary ?

From what I can see, if you change the out-of-the-box default it creates a file to store that setting. A bit of a surprise to me as I thought it would track this on the node type YML file, but clearly not the case as it demonstrates here when you run the export.

DuttonMa:
> > I don't understand why we have all of these 'promote to front page' Yaml files in this change, are they really necessary ?
> 
> From what I can see, if you change the out-of-the-box default it creates a file to store that setting. A bit of a surprise to me as I thought it would track this on the node type YML file, but clearly not the case as it demonstrates here when you run the export.

Yes I see, must just be a peculiarity of this module. Anyway, I just pulled this branch locally and had a play around with it and it seems to work nicely, so all good.

PR #78, opened 2019-10-25
D8 nid 347 publications search

johangant:


config/sync/core.extension.yml
```diff
@@ -110,6 +110,7 @@ module:
   toolbar: 0
   traffic_light_rating: 0
   twig_extensions: 0
+  twig_field_value: 0
```
DuttonMa: Interesting module ! Haven't heard of it before

PR #85, opened 2019-11-13
config changes for D8NID-146, D8NID-155, D8NID-196

David-Corbett:
config changes for block and content type display layouts for contacts, gp practices and health conditions


David-Corbett:
> Looks grand but looks like we might run into problems with this config and the new telephone fields, eg:
> https://github.com/dof-dss/nidirect-drupal/blob/development/config/sync/core.entity_view_display.node.contact.default.yml

I have a solution for that - I've updated it now using the new field type & formatter - you will need to the same for any other 'telephone' fields that you migrating on other content types eg driving instructors, gp practices etc

config/sync/block.block.pagetitle.yml
```diff
@@ -9,7 +9,7 @@ dependencies:
 id: pagetitle
```
johangant: This is interesting to me. Normally I'd expect to see page titles moved between regions using preprocess code; feeding that into a suitable template. Is this wrapped in a block to be easier to incorporate with tools such as Layout Builder or for other reasons?

config/sync/block.block.healthconditionsatoz_2.yml
```diff
@@ -0,0 +1,28 @@
+uuid: 70b5dc81-8444-46e8-800a-19703a0ab734
+langcode: en
+status: true
+dependencies:
+  module:
+    - ctools
+    - nidirect_health_conditions
+  theme:
+    - nicsdru_nidirect_theme
+id: healthconditionsatoz_2
```
johangant: Is there a way to make the `_2` suffix more descriptive? I can't work out what this block does that `block.block.healthconditionsatoz.yml` doesn't provide. If the original doesn't _quite_ do what's needed, maybe we can change it to avoid spawning a new block?

config/sync/block.block.exposedformhealth_conditionssearch_page_2.yml
```diff
@@ -0,0 +1,31 @@
...
+status: true
+dependencies:
+  config:
+    - views.view.health_conditions
+  module:
+    - ctools
+    - views
+  theme:
+    - nicsdru_nidirect_theme
+id: exposedformhealth_conditionssearch_page_2
```
johangant: I would aim to replace the generic `_page_2` suffix with something more descriptive to make it clearer in future what this block filter is for. Probably more relevant for the original view/display definition rather than the exposed form block as that by itself is fairly obvious.

config/sync/block.block.pagetitle.yml
David-Corbett: This is a core block and while it has been placed in a region in the block layout, this is merely to ensure that a title is always rendered out when viewing a given page - as content types have gotten their layouts sorted article, application etc I have been turned this block off, so by the time we get the site fully themed we probably won't need it and it shall be removed

config/sync/block.block.healthconditionsatoz_2.yml
David-Corbett: the `_2` suffix is indicatative that the block appears in the second sidebar region, I can amend the machine name to something more meaningful if you would like. The block is in essence rendered contextually in two regions, if on the search page it (probably) will render on the first sidebar region, while if viewing a health condition page it will render on the second sidebar region. We will also need to amend the original block as well the link txt for resetting the form should only appear when viewing in on a search index page, if viewing on a health condition node page it should not be present

config/sync/block.block.exposedformhealth_conditionssearch_page_2.yml
David-Corbett: same comments apply as above for the `block.block.healthconditionsatoz_2.yml` - the block appears at different places dependent upon the type of page being viewed. It would be sensible to keep the naming of the blocks consist to make it clear what purpose they perform, are you content that the suffix takes the form _side_2?

config/sync/block.block.pagetitle.yml
johangant: Okay, good to know. Thanks!

config/sync/block.block.healthconditionsatoz_2.yml
johangant: Thanks, that explains it. Probably wise to give the filename/machine name a tweak or add a comment to indicate what you've outlined above, but if it's a massive PITA this shouldn't be a show stopper.

David-Corbett: I've updated the 2 health condition blocks tat are output on the 1st and 2nd sidebar regions so that the second sidebar block are suffixed with _sidebar_2 as the machine id

PR #86, opened 2019-11-13
D8NID-350 add 'what links here' feature

johangant:
There's no way in D8 to find out/show what other content links to the node presently being viewed. To that end, I wrote a module to do it for just nodes on three key field types. That's now at https://www.drupal.org/project/whatlinkshere

This PR provides the composer changes, and Drupal config to enable it.


johangant:
> Looks good. Well done for releasing the contrib module. Just one comment on it though, I think that you could make it clearer to the user where the report can be found. We know that it is on the node, but they might look in the 'reports' or 'configuration' sections maybe ?

Good idea, I'll pop that into the contrib module

PR #94, opened 2019-11-20
D8 nid 349 part5

DuttonMa:
Created a new 'NICS Editorial' workflow set in line with other sites installed from the Origins profile (NIDirect was using the standard 'Editorial' workflow set but it isn't possible to configure this in Origins, so a better option to make all sites use the custom 'NICS Editorial' set)


DuttonMa:
Good idea, have removed the 'editorial' workflow set altogether

PR #103, opened 2019-11-26
Site subtopics

DuttonMa:
This view makes up the lower part of the page when you click on a taxonomy term like 'Motoring' in D7.
To get it working I needed to install the views_contextual_filters_or module.
This view will not actually be displayed anywhere as yet, I will handle that when I get into the 'Articles by Term' view that is coming next !

### DIFF
--- a/config/sync/views.view.site_subtopics.yml
+++ b/config/sync/views.view.site_subtopics.yml
@@ -263,14 +263,15 @@ display:
         - url
         - user.permissions
       tags: {  }
-  block_1:
+  site_subtopics_block:
     display_plugin: block
-    id: block_1
+    id: site_subtopics_block
     display_title: 'By Topic'
     position: 1
     display_options:
       display_extenders: {  }
       display_description: ''
+      block_description: 'Site Subtopics'
     cache_metadata:
       max-age: -1
       contexts:
--- a/config/sync/views.view.site_subtopics.yml
+++ b/config/sync/views.view.site_subtopics.yml
@@ -37,6 +37,7 @@ display:
           replica: false
           query_comment: ''
           query_tags: {  }
+          contextual_filters_or: 1
       exposed_form:
         type: basic
         options:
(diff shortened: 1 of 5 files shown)


config/sync/views.view.site_subtopics.yml
```diff
@@ -0,0 +1,281 @@
...
+      max-age: -1
+      contexts:
+        - 'languages:language_content'
+        - 'languages:language_interface'
+        - url
+        - user.permissions
+      tags: {  }
+  block_1:
+    display_plugin: block
+    id: block_1
```
omahm: Use a more descriptive machine name.

config/sync/views.view.site_subtopics.yml
```diff
@@ -0,0 +1,281 @@
...
+      query:
+        type: views_query
+        options:
+          disable_sql_rewrite: false
+          distinct: false
+          replica: false
+          query_comment: ''
+          query_tags: {  }
+          contextual_filters_or: 1
+      exposed_form:
```
johangant: How does the exposed filter block fit into the use case for this view? Or is this boilerplate output that I'd misinterpreted? 😛

config/sync/views.view.site_subtopics.yml
DuttonMa: Yes, have done that now. tbh I hadn't actually decided yet whether to use a block or an 'embed', that will come in phase 2.

config/sync/views.view.site_subtopics.yml
DuttonMa: There are two contextual filters for this view, both of which pull their import from the URL:
- 'parent' which checks the immediate parent of this term in the taxonomy hierarchy
- 'supplementary parent' which checks whether this term has other 'parents' in addition to the term above it in the hierarchy

'Views contextual filters OR' allows rows to be displayed by this view if they match either of the contextual filters above.

johangant: 👍

PR #110, opened 2019-12-05
Block config for footer menu and origins theme settings values

johangant:
NB: To avoid any confusion on first import... the footer menu will likely be empty on initial config import as menu links are considered content and aren't carried with the config export. Once the values are filled in by an admin, or we've faked them in code (if needed) things ought to look clearer :)

### DIFF
--- a/config/sync/block.block.footer.yml
+++ b/config/sync/block.block.footer.yml
@@ -1,0 +1,30 @@
+uuid: 2b367cf7-b333-427d-9bee-920df8e911f8
+langcode: en
+status: true
+dependencies:
+  config:
+    - system.menu.footer
+  module:
+    - system
+  theme:
+    - nicsdru_nidirect_theme
+id: footer
+theme: nicsdru_nidirect_theme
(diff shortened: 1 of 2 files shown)


johangant:
> The social link URLs are nidirect specific - does that matter if these are to be set by admin in the UI? Or should these be left blank in the origin theme settings yml?

I think this is okay, because the config provides the values for these links within the NIDirect site (this repo). If I'd put them into the nicsdru_origins_theme repo that would have been annoying for anyone else using the repo in their project.

The values can be set by a site admin via the theme settings form. The idea being you can change them without developer involvement..... _realises something!_

Okay, so if this PR goes in then any values the admin sets will be lost on the next config import unless those values are exported to code **first**. That makes the process need a developer.

@neilblair if you're happy with the idea, I'll add origins themes social links to the list of excluded config as part of the site launch/launch prep process. That'll keep things locked in for now but mean you can have admins tweak the values in future without losing them to their original values.

PR #111, opened 2019-12-06
Two columns for publish status and moderation status

johangant:


johangant:
Closing this in light of feedback from @neilblair on how the current sites present this.

PR #127, opened 2019-12-12
 D8NID-373 Add related sites menu block to footer region

neilblair:
Creates a menu block for footer "related sites" and adds it to the footer region.


omahm:
Merging as it's pretty old (25 days)

PR #128, opened 2019-12-12
Remove "No front page content has been created yet" and subscribe to …

neilblair:
…links from frontpage view


omahm:
Merging as it's approved and pretty old.

PR #142, opened 2020-01-01
D8 nid 177

omahm:


johangant:
So much YML :)

Looks okay to me at face value, so will approve.

johangant:
There's a config import error from CircleCI to resolve first, obvs, but hopefully that's not too awkward to do.

omahm:
> There's a config import error from CircleCI to resolve first, obvs, but hopefully that's not too awkward to do.

Webform has issues exporting single quotes in HTML textareas. I will escape it in the yaml file.

PR #148, opened 2020-01-03
Recipe missing image

DuttonMa:
Show image on recipe view page, also hide 'page title' block as was breaking layout.
Minimal re-arrangement of fields to make recipe view more like the production page.

### DIFF
--- a/config/sync/block.block.pagetitle.yml
+++ b/config/sync/block.block.pagetitle.yml
@@ -23,6 +23,7 @@ visibility:
     bundles:
       application: application
       article: article
+      recipe: recipe
     negate: true
     context_mapping:
       node: '@node.node_route_context:node'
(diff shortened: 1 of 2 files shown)


config/sync/block.block.pagetitle.yml
```diff
@@ -23,6 +23,7 @@ visibility:
     bundles:
       application: application
       article: article
+      recipe: recipe
```
johangant: Are we still using this? Not a problem if we are, I got the impression from past discussions this was causing some headaches and was being removed in favour of a `$title` variable to be rendered in a given template.

omahm:
I didn't see this and enabled added the image field to the full display mode for recipes in my PR to update the traffic light rating display

DuttonMa:
Ah, looks like this is now obsolete, will bin it !

PR #150, opened 2020-01-06
Turn off this block rendering on nodes

johangant:
- We can use {{ drupal_block('plugin_id') }} in specific templates, where the page title needs to sit inside a content (or other) region


johangant:
Matching up with PR https://github.com/dof-dss/nicsdru_nidirect_theme/pull/31

PR #153, opened 2020-01-07
D8 nid 223

omahm:


config/sync/core.extension.yml
```diff
@@ -83,7 +84,6 @@ module:
   nidirect_hospital_waiting_times: 0
   nidirect_money_advice_articles: 0
   nidirect_news: 0
-  nidirect_webforms: 0
```
johangant: Is this module removal expected?

omahm: Hell no, and it's enabled on my site... odd.

PR #154, opened 2020-01-09
D8NID-399 rename toc field to match origins module

johangant:
D7 field was called field_enable_toc, Origins module that NIDirect D8 uses calls this field_toc_enable.

This PR contains config changes for field instance/storage, entity types and block config to present the ToC block


johangant:
NB: this is potentially breaking PR. At the very least, you will need to rollback all migrations before applying new config, then re-migrate when done.

Probably easiest to reinstall from scratch using the present site config, then migrate content in again. This ensures the post-migrate events trigger as expected; they're generally considered one way streets in terms of the effects of their data changes and there's not sufficient value to write rollback events to reverse as this is an irregular task and won't be needed once the site is live.

PR #168, opened 2020-01-16
Render the page title block on taxonomy term pages

johangant:
- Dovetails with block render on theme template for these pages.


johangant:
See https://github.com/dof-dss/nicsdru_nidirect_theme/pull/39 for reference

PR #181, opened 2020-01-29
permission changes for workflow

DuttonMa:
Most roles were lacking workflow permissions and were unable to manipulate content at all


neilblair:
This can be merged - but Circle CI is complaining - something about missing spaces ... Ignore and merge anyway?

PR #188, opened 2020-02-03
D8 nid 222 1

omahm:


omahm:
I have my local GA API settings in here for the moment, we will need to switch these out on dev & prod.

omahm:
> should the client_id and default_config_hash in file config/sync/google_analytics_counter.settings.yml be masked or omitted from config and the values for these kept only the DB - thinking that these are hashes for connecting to the Google API's thesefore probably don't want these in a public git repo

Aye, that's what I was referring to n the previous comment. I'm not worried about the keys as it's my  GA ID and limited to my local URL.

composer.json
```diff
@@ -195,6 +196,9 @@
             },
             "drupal/easy_install": {
                 "Handle missing core version": "https://www.drupal.org/files/issues/2019-12-19/easy_install_core_version-3101883-0.patch"
+            },
+            "drupal/google_analytics_counter": {
+                "Add API query tweak options": "https://gist.githubusercontent.com/omahm/02ac3b7138c4d0518bd954721b74b763/raw/8d7ffa08b21359c7bf11165cdef1fb7e98483774/gac_query_options.patch"
```
johangant: If this is re-usable, might be worth adding a new drupal.org issue + attach the patch there?

omahm:
Removing API keys to stop GG from nagging.

composer.json
omahm: Yup, added it as a feature request yesterday https://www.drupal.org/project/google_analytics_counter/issues/3110812

PR #192, opened 2020-02-06
config amendments for search views to ensure correct output of code

David-Corbett:


config/sync/views.view.search.yml
```diff
@@ -373,7 +377,7 @@ display:
           group_type: group
           admin_label: ''
           empty: false
-          content: '@total results'
+          content: '<h2 class="view--count">@total results</h2>'
```
omahm: Although the chances are slim, do we want to singular text in the header?

David-Corbett: possibly? All I'm doing is wrapping the result with a h2 tag to ensure that it works properly on the page - someone else configured the output for inside the block. I can add a ticket in to get that amended and bring it in line with the other searches

omahm: Yeah, just spotted that it would have that support tho. I'm happy enough to do it as part of the next sprint.

PR #201, opened 2020-02-10
D8 nid 264

DuttonMa:
Permissions adjustments, mostly focussed on making sure that user roles can create the content that they need to.  Changes also made to allow most roles to view GPs.


johangant:
Thanks. This looks fairly self-explanatory to me and I would imagine it mirrors the D7 site's functionality. I'd recommend a sense check from @neilblair for advancing this, a thumbs up from me is probably going to give far less confidence and take longer :)

PR #206, opened 2020-02-11
Fix proximity sort

johangant:
- Need to force field output because views uses this as part of
  its select fields, enabling the sort to then function later on.
- Without this, you're effectively selecting NID (for rendering)
  then trying to sort on a field/computed value that you don't have
  and the sort operation fails.


johangant:
Forcing the change through due to unrelated coding standards fails in the theme repo project.

PR #213, opened 2020-02-14
D8 nid 476 search api spellcheck

johangant:
NB: I expect this will grumble because the PR at https://github.com/dof-dss/nidirect-site-modules/pull/110 hasn't merged. Ignore the config import error in Circle CI until then, kthxbai 😁


johangant:
Circle CI won't let this pass because the module to install isn't available in the release specified by this version of the composer.lock file.

Rather than cut a new release tag in the nidirect-site-modules repo and update the lock file here, I'm inclined to park this until we're ready to merge more stuff from the sprint and let the test team have this as one discrete release.

PR #216, opened 2020-02-14
D8NID-220

David-Corbett:
adds a new content type of 'feature' and tweaks various views & blocks etc to get stuff to render properly on the homepage


omahm:
I would explore the possibility of changing this over to use a 'featured' blocks/paragraphs as it isn't linked to existing content types anymore and feels a bit overkill. This would also facilitate the removal of the entity queue module.

DuttonMa:
Just to throw my opinion in here I would think that using paragraphs would be the most efficient way to achieve the desired result and would also be less confusing for content editors.

johangant:
So to summarise what's happening here:

- Entityqueue had to come out, because it did not meet the editorial requirements around revisioning and scheduled publishing.
- To replace it, a 'Featured content list' content type is defined to present a consistent set of features and layout around scheduled publishing, moderation and the use of entity reference fields allow content simple re-ordering. It contains references of up to 6 feature nodes and can be reused across the site in future when needed.
- The default frontpage view has a Page display, but we don't use it anymore so it's gone.

There is a PR for the custom code supporting all this at: https://github.com/dof-dss/nidirect-site-modules/pull/138. That provides a replacement for the view listed above as well as a block plugin to 'inject' a featured content list node into the page bottom region (config for that in here).

johangant:
Oh and the config import will fail until the custom code repo PR is merged and composer is updated here to reference the latest release including the nidirect_homepage module.

PR #221, opened 2020-02-19
D8 nid 514

omahm:


omahm:
> Looks good, but should there be a composer change in here too for the traffic light module ?

Yup, good catch!

PR #224, opened 2020-02-20
D8NID-502

David-Corbett:
amendments to the gp practices proximity view to ensure correct markup is being rendered


config/sync/views.view.gp_practices_proximity.yml
```diff
@@ -71,14 +71,12 @@ display:
...
         options:
-          row_class: ''
+          grouping: {  }
+          row_class: card
           default_row_class: false
-          uses_fields: true
-          type: ul
-          wrapper_class: item-list
-          class: ''
+          uses_fields: false
```
omahm: Are you sure this should be false? 
CC @johangant

David-Corbett: pretty sure! Might be changed later, but for now it needs to match the other view

johangant: I think it's fine to change the format options to unstyled list. The key thing that needs to be enabled is the 'Force using fields' option as without it, the proximity search query fails.

johangant: Hard to say whether that option maps to the 'uses_fields' directive in there. The general idea is you want to show Content (rather than individual fields) in a specific view mode, and you want that obscure 'Force fields' option enabled to ensure the query is well-formed.

David-Corbett:
> @David-Corbett marked the earlier conversation about the 'force fields' views display config option as resolved, so I'm happy to approve this on the basis that this option has been checked to still be enabled.

Done!

PR #225, opened 2020-02-20
D8 nid 515

omahm:


config/sync/pathauto.pattern.terms_default.yml
```diff
@@ -10,25 +10,19 @@ label: 'Terms - default'
...
-      accessni_ub_sectors: accessni_ub_sectors
-      accessni_ub_services: accessni_ub_services
-      districts_in_northern_ireland: districts_in_northern_ireland
       drive_instr_categories: drive_instr_categories
       hc_body_location: hc_body_location
       hc_body_system: hc_body_system
       hc_condition_type: hc_condition_type
       hc_info_sources: hc_info_sources
       hc_symptoms: hc_symptoms
-      recipe_course_types: recipe_course_types
```
johangant: Not strictly needed but figured because you had it removed from the other task + then ran config export it'll also be removed here. They're both coming out anyway so no big deal

omahm:
> Looks good to me. The recipe path aliases config shouldn't strictly be in the diff here but because it's part of the same overall piece of work it's no problem.

Yeah I jumped the gun a little on 515 before 514 was merged in so had to deal with a few merge conflicts

PR #230, opened 2020-02-25
Flushed config for content types

johangant:
Dropped origins LB search + TOC from landing page


johangant:
Most of this is boilerplate config export. Worth some extra eyes on this in case it's missed or exported anything unsual.

PR #231, opened 2020-02-25
Execute build + kernel tests nightly Mon-Fri

johangant:


.circleci/config.yml
```diff
@@ -244,3 +244,17 @@ workflows:
       - unit_kernel_tests:
           requires:
             - build
+  nightly:
+    triggers:
+      - schedule:
+          # At 00:00 on every day-of-week from Monday through Friday
```
omahm: Do you crontab.guru much? 😄

johangant: The choice of champions 😉

PR #236, opened 2020-02-27
D8 nid 468 password policy

johangant:
Contains policies to match existing NICS policy


neilblair:
@johangant - sorry took me a while to look at this.  Good to go - but needs composer.lock recreated

johangant:
Thanks. Sorry, without the context of the JIRA ticket this looks like I've left it dangling. What I should have put is that we can refresh this/merge it once we bring that item of work into a sprint or when capacity + priorities allow it. I'll hold off merging for the moment :)

PR #238, opened 2020-03-02
D8NID-571 Add job for functional tests with JS on config only site

johangant:
NB: Won't fully work until https://github.com/dof-dss/nidirect-site-modules/pull/123/ is merged and a new release is created.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -244,3 +244,17 @@ workflows:
       - unit_kernel_tests:
           requires:
             - build
+  nightly:
+    triggers:
+      - schedule:
+          # At 00:00 on every day-of-week from Monday through Friday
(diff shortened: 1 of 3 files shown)


johangant:
You'll note that the DB namespace env var is temporarily replaced with a literal value because it was taking absolutely ages to work around bash/Apache escaping the backslash characters. There's nothing super-secret about our choice of DB vendor or the class we use for it, to my knowledge, but happy to swing back to that point between sprints.

johangant:
So this does work now with 0.1.7 of nidirect-site-modules, but it's still hitting some of the intermittent web installer issues presently being found in Drupal core. I think it's safe to merge now providing we don't hang our PR merging ability off that one condition right now.

PR #249, opened 2020-03-12
D8nid 565

neilblair:
Some config changes in relation to revision comparison - a number of disabled fields have been added to the default display mode for various content types to enable authors to see them when comparing revisions.  I note there is a revision comparison display mode which can be enabled for content types - but I decided it was easier to just modify the default display mode because that is the initial display mode authors will see when they do a revision comparison ...


config/sync/core.entity_view_display.node.driving_instructor.default.yml
```diff
@@ -31,58 +91,56 @@ content:
...
     region: content
-    label: above
+    label: inline
     settings:
       link: true
     third_party_settings: {  }
   field_di_categories:
-    type: html_list_formatter
-    weight: 6
+    type: entity_reference_label
```
omahm: Why was this changed from html_list_formatter?

config/sync/core.entity_view_display.node.article.default.yml
```diff
@@ -28,108 +54,93 @@ bundle: article
...
     label: above
     settings: {  }
     third_party_settings: {  }
-    type: metatag_empty_formatter
-    region: content
   field_photo:
-    type: entity_reference_entity_view
-    weight: 14
-    label: above
+    type: media_thumbnail
```
omahm: @johangant is this undoing the changes you did recently?

config/sync/core.entity_view_display.node.news.default.yml
```diff
@@ -23,39 +48,67 @@ content:
...
     weight: 3
-    label: above
+    label: hidden
     settings:
-      link: true
+      image_style: ''
+      image_link: ''
     third_party_settings: {  }
-    type: entity_reference_label
+    type: media_thumbnail
```
omahm: Review this type

config/sync/core.entity_view_display.node.driving_instructor.default.yml
neilblair: That's an oopsy - will change that back

johangant:
> Some config changes in relation to revision comparison - a number of disabled fields have been added to the default display mode for various content types to enable authors to see them when comparing revisions. I note there is a revision comparison display mode which can be enabled for content types - but I decided it was easier to just modify the default display mode because that is the initial display mode authors will see when they do a revision comparison ...

Adding a view mode isn't a problem and I think it would definitely make sense here. For example: I had considered using the default view mode for entity embed previews, but that would then clash with this work :)

Also, adding new modules and functionality can add fields or changes to that default view mode so I'd 💯 avoid using it to keep a clear separation.

config/sync/core.entity_view_display.node.driving_instructor.default.yml
omahm: No worries... things are changing that quickly with the project it's bound to happen.

config/sync/core.entity_view_display.node.article.default.yml
johangant: @omahm it's the default view mode so shouldn't matter. I'd be inclined to move this to a new view mode called 'diff', or something like that for reasons I've outlined in a PR comment below.

neilblair: I added the Photo field into the default display mode for revision comparison purposes.  However the image style should be 'None (original image)' - not media thumbnail.  We'll need to review all image styles and apply them appropriately at a later stage (theres a task in backlog for that).

neilblair:
Ok I'll look at enable the revision comparison view mode and maybe a form alter on diff comparison page controls for setting the view mode 👍

johangant:
Good stuff. When done, click the little curvy arrows up near the reviewers list and that'll ping me to come look at this again 👍

neilblair:
Closed this PR - see #252 instead

PR #252, opened 2020-03-13
Adds config for revision comparison display mode

neilblair:
This PR replaces [https://github.com/dof-dss/nidirect-drupal/pull/249](https://github.com/dof-dss/nidirect-drupal/pull/249) and creates a separate revision comparison display mode for relevant content types.


config/sync/core.entity_view_display.node.driving_instructor.diff.yml
```diff
@@ -0,0 +1,107 @@
...
+  field_di_areas:
+    type: entity_reference_label
+    weight: 8
+    region: content
+    label: inline
+    settings:
+      link: false
+    third_party_settings: {  }
+  field_di_categories:
+    type: entity_reference_label
```
johangant: Does this still need changing back? See https://github.com/dof-dss/nidirect-drupal/pull/249#discussion_r392134139

config/sync/core.entity_view_display.node.news.diff.yml
```diff
@@ -0,0 +1,123 @@
...
+    settings: {  }
+    third_party_settings: {  }
+  field_photo:
+    weight: 4
+    label: hidden
+    settings:
+      image_style: ''
+      image_link: ''
+    third_party_settings: {  }
+    type: media_thumbnail
```
johangant: @omahm left a question about this type in https://github.com/dof-dss/nidirect-drupal/pull/249#discussion_r392096314 - does it need to be `entity_reference_label` or okay as `media_thumbnail`?

config/sync/core.entity_view_display.node.driving_instructor.diff.yml
omahm: If you want it consistent with the full display, yes this would need to be changed.

neilblair: The html_list_formatter is outputting the list of categories without a label. This is perfect for the teaser display (the full display isn't really being used as driving instructor nodes are only viewed in driving instructor search results using the teaser display) - but it looks incongruous with the output of the other fields in the diff display which are output with a label and output is controlled with field.html.twig and diff display specific field--node--diff.html.twig templates.

config/sync/core.entity_view_display.node.news.diff.yml
neilblair: I think media_thumbnail is OK for now - but later it will need to change to a responsive image style for news images.

PR #259, opened 2020-03-18
D8NID-576 platform.sh config

johangant:
PR to:

- Define the required services and config for platform.sh environments.
- Solr configuration to work around some quirks of our setup and theirs.
- Added fast.ly module for use whenever it's needed in future.
- Split of services and settings to allow us to easily swap in/out rather than dynamically rewriting segments of general equivalent files.
- Config split for production environment.
- Define config for Redis service


johangant:
I _think_ I've finished spamming the commits now to get around awkward things like Solr config, so happy to put this forward for review. There's a discussion to be had tomorrow for any smaller points around files added to the web folder but otherwise, this all appears to work smoothly on platform.sh

johangant:
Caveat relating to Circle CI: we're still stuck with a purely CI service there because the deploy step was getting intermittent SSH key rejections from platform.sh for reasons unknown. That's being followed up by their team so, for now, we have a basic GitHub webhook integration service which is pretty good but potentially means we can ship stuff we wouldn't want to, if we're careless.

DuttonMa:
Those "intermittent SSH key rejections from platform.sh" sound exactly like the ones that I have been experiencing locally, let's hope they get sorted out soon.

.platform.app.yaml
```diff
@@ -0,0 +1,133 @@
...
+
+      rules:
+        # Provide a longer TTL (2 weeks) for aggregated CSS and JS files.
+        '^/sites/default/files/(css|js)':
+          expires: 2w
+
+# The configuration of scheduled execution.
+crons:
+  drupal:
+    spec: '*/20 * * * *'
```
neilblair: TODO: run cron every minute to ensure content schedules run on time

.platform/solr_config/stopwords_en.txt
```diff
@@ -0,0 +1,35 @@
...
+for
+if
+in
+into
+is
+it
+no
+not
+of
+on
```
neilblair: TODO: compare with D7 solr config stopwords.  Defo need to remove 'on' so that searches for "go on ni" work.

PR #261, opened 2020-03-19
D8NID-563 ~ Update Core to 8.8.3

omahm:


composer.json
```diff
@@ -97,6 +97,7 @@
         "drupal/linkit": "^4.3",
         "drupal/media_browser_responsive_images": ">1.x-dev",
         "drupal/media_entity_browser": "^2.0",
+        "drupal/memcache": "^2.0",
```
DuttonMa: Adding memcache back in, is this accidental ?

omahm: Keeping it in as some people will get complains around it. Will eventually remove it.

PR #265, opened 2020-03-20
D8NID-592 ~ Purge composer of redundant config

omahm:


composer.json
```diff
@@ -1,14 +1,13 @@
 {
-    "name": "nics/nidirect",
```
johangant: Do we want to keep any custom metadata like this for the project? Probably not terribly important right now but it's not clear whether this is tied at all to any specific composer behaviour or whether it's purely information and can be changed at any time.

composer.json
```diff
@@ -222,47 +175,46 @@
...
-            "drush/Commands/{$name}": [
-                "type:drupal-drush"
-            ]
+            "web/modules/custom": ["dof-dss/nidirect-site-modules"],
+            "web/modules/origins": ["dof-dss/nicsdru_origins_modules"],
+            "web/modules/migrate/{$name}": ["dof-dss/nidirect-migrations"],
+            "web/modules/contrib/{$name}": ["type:drupal-module"],
+            "web/modules/custom/{$name}": ["type:drupal-custom-module"],
+            "web/themes/custom/{$name}": ["type:drupal-custom-theme"],
+            "web/themes/contrib/{$name}": ["type:drupal-theme"],
```
DuttonMa: I can't quite see how the theme repos will be pulled in, as they have been deleted from the top of this file  and I can't now see a link to dof-dss/nicsdru_nidirect_theme.git or git@github.com:dof-dss/nicsdru_origins_theme.git anywhere in this file ?

omahm: The themes are included as packages on lines 39 and 41

composer.json
omahm: Left this in as a quick reference back to the origin of the composer file and that we are no longer using the old drupal-project github project.

composer.json
DuttonMa: Ah yes I see, the diff was confusing me !

PR #274, opened 2020-03-26
D8 nid 513

omahm:


omahm:
> Are the landing page changes expected? There's a full view mode in there for it. If it's useful let's keep it, but it confused me a little with the scope of the PR at first

I'm still not 100% what direction they want to go with this but I do know I want to rip out the old field indexing solution. Added the Full Content display as it's a bit odd not to have it given the node will be viewed. It would be different if it was for say Driving Instructors where the node is never viewed.

PR #278, opened 2020-03-31
Safely handle config_readonly mode on platform.sh

johangant:
Config_readonly is a great module, it stops admins making changes to Drupal configuration through the UI that will be overwritten on the next deployment; a very frustrating situation for all.

However, setting the value to be active via an environment variable means when platform.sh deploys it can't apply database or config updates and leaves it in a rather half-baked state.

These changes will:

1. Default config_readonly to ACTIVE if there is no preference set already.
2. Disable it in the event the PHP_SAPI_CLIENT is present; ie platform.sh or drush.

I've also sneaked in some environment indicator colours for prod env, because I had it open in front of me :)

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -54,16 +54,18 @@ hooks:
     set -e
     bash install-redis.sh 5.1.1
     curl -sS https://platform.sh/cli/installer | php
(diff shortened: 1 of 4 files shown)


johangant:
> LG, take a look at the null coalesce operator as a nice replacement for !empty() statements

Thanks! I had some bother with them in the past as they're my normal approach now. Can't remember what was the exact trouble before. I'll get this merged in for now and the next time I work near here I'll re-test/adjust 👍

PR #284, opened 2020-04-03
D8 nid 619

DuttonMa:
- Expand 'General' and 'Detail' field groups by default
- Remove table of contents from Contacts


config/sync/core.entity_view_display.node.contact.default.yml
```diff
@@ -107,21 +106,16 @@ content:
...
-      format_custom_true: ''
-    third_party_settings: {  }
-    type: boolean
-    region: content
   links:
     weight: 1
     region: content
     settings: {  }
     third_party_settings: {  }
+  toc_display:
```
omahm, line 114: Does this field not have to come out?

config/sync/core.entity_view_display.node.contact.full.yml
```diff
@@ -160,6 +159,11 @@ content:
     region: content
     settings: {  }
     third_party_settings: {  }
+  toc_display:
```
omahm, line 162: Same here, I don't think we need the toc_display , they should be in the hidden config.

PR #288, opened 2020-04-08
D8NID-476 solr suggest handlers

johangant:
Adds a variety of Solr config changes sourced from the D7 site. Not all the detail for each core has been carried across, these will be added as separate core configurations as and when required after this PR is in.


.platform.app.yaml
```diff
@@ -123,5 +123,5 @@ web:
 # The configuration of scheduled execution.
 crons:
   drupal:
-    spec: '* * * * *'
+    spec: '*/5 * * * *'
```
johangant, line 126: Set to 5 mins rather than every minute, because platform.sh only permits crons to run a minimum of every 5 minutes, even if you set it to less than that.

.platform/services.yaml
```diff
@@ -17,6 +17,7 @@ db:
         privileges:
           main: admin
 
+# Fudging the service name forces a full rebuild with any new config.
```
johangant, line 20: Left as a reminder

johangant:
Adding reviewers early on to this PR (apologies for any commit spam that might follow!).

As per the initial comment, this is intended to give a big step towards the present D7 behaviour but not expected to be the final set of changes.

.platform/solr_config/solrconfig_extra.xml
```diff
@@ -1,188 +1,190 @@
...
-  Special configs for Term Vector
-  7.0.0
--->
-  <searchComponent name="tvComponent" class="solr.TermVectorComponent"/>
+  <!--
+    Special configs for Elevator
+    7.0.0
+  -->
+<searchComponent name="elevator" class="solr.QueryElevationComponent">
+<str name="queryFieldType">string</str>
```
neilblair, line 183: Can this queryFieldType be changed from string to text_ws?

When set to "string", the elevator component doesn't do any special processing of the query terms before comparing them with elevate.xml entries.

So if I search for "Go ON NI", the elevator component will not elevate the entry for "go on ni" because of the case difference between the search terms and query text specified for the elevate entry.

The text_ws field type does some tokenisation and converts query terms to lower case.  Seems to be present in 7.x schema.xml so should be safe to use it.

```
    <!-- A text field that only splits on whitespace for exact matching of words -->
    <fieldType name="text_ws" class="solr.TextField" omitNorms="true" positionIncrementGap="100">
      <analyzer>
        <tokenizer class="solr.WhitespaceTokenizerFactory"/>
        <filter class="solr.LowerCaseFilterFactory"/>
      </analyzer>
    </fieldType>
```

johangant: Yes, you're quite right here. I've got this open in front of me today after I had the extra information through from yourself and @David-Corbett yesterday. No hurry to commit this at the moment so I'll evaluate and commit changes, then re-request a review once it looks about right.

johangant:
Closing this PR off as it references the wrong ticket. I'll raise a new Draft PR and cross-link to here.

PR #290, opened 2020-04-08
D8NID-555

DuttonMa:
add and configure new contact child view/block


config/sync/views.view.contact_children.yml
```diff
@@ -0,0 +1,238 @@
...
+    cache_metadata:
+      max-age: -1
+      contexts:
+        - 'languages:language_content'
+        - 'languages:language_interface'
+        - url
+        - 'user.node_grants:view'
+        - user.permissions
+      tags: {  }
+  block_1:
```
johangant, line 222: Worth changing this display ID to something more descriptive? It's a pattern we've tried to go for elsewhere in the project so it's easier to see what you're dealing with in config or custom code.

DuttonMa: Good spot - that's done now

PR #294, opened 2020-04-14
D8 nid 610

omahm:


config/sync/core.extension.yml
```diff
@@ -73,7 +73,29 @@ module:
   metatag_twitter_cards: 0
   metatag_verification: 0
   migrate: 0
+  migrate_drupal: 0
+  migrate_nidirect_file: 0
```
johangant, line 77: You probably don't want all these migration extensions unless something's changed in the project requirements. My guess is they're enabled locally for something else and haven't been excluded here

omahm: ahh my bad..  I'll clear that up

PR #301, opened 2020-04-28
D8 nid 270

omahm:


omahm:
>
> The other tours may be useful for some developers, but I don't think that we really need them.
> 

They are the defaults that are installed when you enable the Core Tour module.

DuttonMa:
Yes I realised that these tours were the defaults that come with the Tour module, the point that I was making is that one of these tours covers the admin/content admin screen and that we have made so many changes to it that this tour seems out of date and not very relevant. I was thinking that we should either not have it at all or, if we do have it, it should be updated to reflect the changes that we have made.

omahm:
> Yes I realised that these tours were the defaults that come with the Tour module, the point that I was making is that one of these tours covers the admin/content admin screen and that we have made so many changes to it that this tour seems out of date and not very relevant. I was thinking that we should either not have it at all or, if we do have it, it should be updated to reflect the changes that we have made.

The admin/content tour is the one I created yesterday using the latest build of the site, what has changed since then?

PR #305, opened 2020-05-04
D8NID-621

DuttonMa:
Allow unpublish to draft


config/sync/core.entity_form_display.node.feature.default.yml
```diff
@@ -88,6 +88,12 @@ content:
     region: content
     settings: {  }
     third_party_settings: {  }
+  publish_state:
```
johangant, line 91: Not sure if these options are needed on the feature node type but guess they don't do any harm as they stand

PR #308, opened 2020-05-04
D8 nid 607

neilblair:
Adds:
- basic image styles used across content types
- responsive image styles
- makes relevant responsive styles available in CKEditor
- news thumbnail display mode for media image entities
- alters the news teaser display to use the news thumbnail media display mode

TODO: need to look at mapping image styles from D7 to D8 in migration scripts. I was unable to match D7 machine names for image and responsive image styles in the D8 equivalents 😒


web/.editorconfig
```diff
@@ -0,0 +1,17 @@
+# Drupal editor configuration normalization
```
johangant, line 1: Is this file needed? Looks like it was duplicated from the directory above `/web`?

web/.gitattributes
```diff
@@ -0,0 +1,61 @@
+# Drupal git normalization
```
johangant, line 1: Is this file needed? Looks like it was duplicated from the directory above `/web`?

web/.editorconfig
neilblair: Yep not sure how that happened - removed it

web/.gitattributes
neilblair: Ditto - removed this one too

neilblair:
Removed both of those files Johan

neilblair:
Just added another commit - a tweak to image styles intended for news thumbnails - renamed them to be more generic "nigov landscape thumbnail" for better reuse.

neilblair:
...And another commit to fix the new thumbnail responsive image style which got broken by the previous commit!

PR #310, opened 2020-05-05
Add placeholder field to media listing

johangant:
Preprocessed from nidirect_media.module

### DIFF
--- a/config/sync/views.view.media.yml
+++ b/config/sync/views.view.media.yml
@@ -29,8 +29,12 @@ display:
         options:
           perm: 'access media overview'
       cache:
-        type: tag
-        options: {  }
+        type: time
+        options:
+          results_lifespan: 60
+          results_lifespan_custom: 0
+          output_lifespan: 60
+          output_lifespan_custom: 0
       query:
         type: views_query
         options:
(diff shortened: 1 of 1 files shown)


config/sync/views.view.media.yml
```diff
@@ -29,8 +29,12 @@ display:
         options:
           perm: 'access media overview'
       cache:
-        type: tag
-        options: {  }
+        type: time
```
johangant, line 32: Set to time based because if you reference a media item from another place, it doesn't automatically pop the cache item for the view row in the result. It's fairly complex to add these dependencies so a minute's query and render cache TTL is used here to provide a reasonable compromise for now.

config/sync/views.view.media.yml
```diff
@@ -499,6 +503,55 @@ display:
...
+          table: views
+          field: nothing
+          relationship: none
+          group_type: group
+          admin_label: ''
+          label: 'Referenced from'
+          exclude: false
+          alter:
+            alter_text: true
+            text: ''
```
johangant, line 517: This is added from nidirect.module's preprocess hook for views fields.

PR #311, opened 2020-05-06
D8 nid 607

neilblair:
Creates and configures new media entity view modes needed to apply responsive image styles to media entity reference image fields (have assumed here that images in landing pages could be media entity reference fields).

Also changed the photo field on feature teaser display (/admin/structure/types/manage/feature/display/teaser) to use the new feature card display mode.

And also configured the photo field in health conditions full display mode to use the existing embed display mode.


neilblair:
Wondered about that smart_trim myself - not sure where it came from.  Very popular looking at the module page https://www.drupal.org/project/smart_trim

PR #314, opened 2020-05-11
D8NID-555

DuttonMa:
Added contact children view and block

(Replacement for https://github.com/dof-dss/nidirect-drupal/pull/290 as that had got old and stale. Comments from that PR have been incorporated in this change.)


DuttonMa:
> Looks fine. There is no block config to add the views block to a theme region, unless I've missed something or it's to follow later?

That's a good point Johan, and there did used to be one, but now the block is included in a Twig template directly instead, as seen in https://github.com/dof-dss/nicsdru_nidirect_theme/pull/109

PR #321, opened 2020-05-19
D8 nid 659

DuttonMa:
Remove moderation control widget from all displays now that moderation sidebar has been introduced.

### DIFF
--- a/config/sync/core.entity_view_display.node.application.default.yml
+++ b/config/sync/core.entity_view_display.node.application.default.yml
@@ -17,6 +17,7 @@ dependencies:
   module:
     - link
     - metatag
+    - smart_trim
     - text
     - user
 id: node.application.default
@@ -25,32 +26,27 @@ bundle: application
 mode: default
 content:
   banner_display:
-    weight: 0
+    weight: 1
+    region: content
     settings: {  }
     third_party_settings: {  }
-    region: content
   body:
     label: hidden
     type: text_default
-    weight: 2
-    settings: {  }
-    third_party_settings: {  }
-    region: content
-  content_moderation_control:
-    weight: -20
+    weight: 3
     settings: {  }
     third_party_settings: {  }
     region: content
   field_additional_info:
     type: text_default
-    weight: 4
+    weight: 5
     region: content
     label: above
     settings: {  }
     third_party_settings: {  }
   field_link:
     type: link
-    weight: 3
+    weight: 4
     region: content
     label: above
     settings:
@@ -61,14 +57,14 @@ content:
       target: ''
     third_party_settings: {  }
   field_meta_tags:
-    weight: 9
+    weight: 10
     label: above
     settings: {  }
     third_party_settings: {  }
     type: metatag_empty_formatter
     region: content
   field_related_info:
-    weight: 10
+    weight: 11
     label: above
     settings:
       trim_length: 80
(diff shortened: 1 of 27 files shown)


config/sync/core.entity_view_display.node.application.default.yml
```diff
@@ -17,6 +17,7 @@ dependencies:
   module:
     - link
     - metatag
+    - smart_trim
```
johangant, line 20: Is that expected/desired? Looks out of scope of the ticket but could well be what's needed, regardless.

config/sync/core.entity_view_display.node.article.default.yml
```diff
@@ -20,6 +20,7 @@ dependencies:
   module:
     - link
     - metatag
+    - smart_trim
```
johangant, line 23: Is this an expected/wanted addition?

config/sync/core.entity_view_display.node.embargoed_publication.default.yml
```diff
@@ -17,6 +17,7 @@ dependencies:
     - datetime
     - metatag
     - options
+    - smart_trim
```
johangant, line 20: Wanted?

config/sync/core.entity_view_display.node.gp_practice.default.yml
```diff
@@ -89,6 +84,14 @@ content:
         style: ''
         preferScrollingToZooming: 0
         gestureHandling: auto
+      data_provider_settings: {  }
+      centre:
```
johangant, line 88: What's changed to make this entity spit this extra config out here? It might well be harmless but thought It wise to check

config/sync/core.entity_view_display.node.health_condition.default.yml
```diff
@@ -117,9 +113,19 @@ content:
   field_summary:
     weight: 6
     label: above
-    settings: {  }
+    settings:
```
johangant, line 116: Is this autogenerated config? Looks unimportant being on the default display mode but unclear why it's coming out as part of this work/at this time.

config/sync/core.entity_view_display.node.application.default.yml
DuttonMa: Good question, have investigated and this just seems to be an oddity.
If you pull latest development, import config, make any minor change to the Article display and then export, smart_trim appears ! 
I think the question is not "why is it being added " but is really "why wasn't it there already" - as smart_trim is already being used in Article display.

config/sync/core.entity_view_display.node.article.default.yml
DuttonMa: Same answer as above

config/sync/core.entity_view_display.node.gp_practice.default.yml
DuttonMa: I rolled back and made the moderation control change again and these lines were added in the same way. We'll just have to put it down to a config export oddity I think.

johangant: That's okay. It's on the default view mode so shouldn't affect anything visually. Another reason never to use the default view mode for theming related work!

config/sync/core.entity_view_display.node.health_condition.default.yml
DuttonMa: Again, I tried this again with the same end result, certainly doesn't look harmful.

PR #323, opened 2020-05-20
Add help/context to featured content list reference field

johangant:


config/sync/field.field.node.featured_content_list.field_featured_content.yml
```diff
@@ -11,7 +11,7 @@ field_name: field_featured_content
 entity_type: node
 bundle: featured_content_list
 label: 'Featured content'
-description: ''
+description: "<p>Defines the order of feature nodes to appear as part of this featured content list.</p>\r\n<p>Please note: the homepage will only show the first three items defined.</p>"
```
omahm, line 14: Do we want to review to 'nodes' or use something like 'content' ?

johangant: Nodes is technically accurate, but go on then... 'content' is likely more editor-friendly and not confusing to developers :)

PR #332, opened 2020-05-26
D8NID-683 solr config

johangant:
This PR aims to improve Solr spelling suggestions by:

- Splitting content into indexes for general content, contacts and health conditions.
- Adjusting the search API config and Solr config to make suggestions more generous.

Some byproducts of the process include:

- Needed to recreate the contacts and health conditions views config, exposed blocks and autocomplete config.


johangant:
There's also some custom code to make suggestions become links again after the link manager JS tries to convert it to a span. That's in a separate PR.

PR #338, opened 2020-06-01
D8NID 706+678: fix thin banner and mandatory news intro paragraph

johangant:


config/sync/core.entity_view_display.node.article.default.yml
```diff
@@ -53,9 +53,10 @@ content:
     weight: 15
     label: above
     settings:
-      link: true
+      view_mode: banner_thin
```
johangant, line 56: This is functional, but feels like it shouldn't be needed. Look at `nidirect_common_preprocess_page` and you'll see it uses an EntityViewBuilder to render the banner image media entity in defined view mode. You _should_ be able to pass in a view mode such as 'banner_thin' here, but this doesn't seem to work for reasons I can't determine at the moment.

Instead, it seems to look at the referring node and the default view mode and extract the field formatter settings from that (which is set correctly in this PR) and uses that to render the banner image rather than a label with link.

omahm: Interesting quirk that the EVB doesn't respect that view mode.

PR #343, opened 2020-06-02
D8NID-659

DuttonMa:
Remove admin tabs for content editing as now handled by moderation sidebar


config/sync/block.block.tabs.yml
```diff
@@ -1,6 +1,6 @@
 uuid: f568248f-9226-4af2-b819-d236889365a1
 langcode: en
-status: true
+status: false
```
johangant, line 3: Ha, this is so concise and so satisfying 🕺

PR #356, opened 2020-06-11
Solr spellcheck config refinements

johangant:
Removes the search API spellcheck field, because we use the `copyField` feature of Solr to copy all fulltext fields into a multi-value text field (see solr_schema.xml). This is a bit more adept at handling things like referenced entity labels for starters.

Also replicates the processor plugins to whittle down HTML and only index published entities across all indexes.


johangant:
This PR follows the initial feature PR, because on deployment to platform.sh Solr simply wouldn't fetch results and failed silently :(

PR #367, opened 2020-06-26
Sets help text (descriptions) on fields in the article content type.

neilblair:


config/sync/field.field.node.article.field_banner_image.yml
```diff
@@ -11,7 +11,7 @@ field_name: field_banner_image
 entity_type: node
 bundle: article
 label: 'Banner image'
-description: ''
+description: 'Optional thin banner image (1880 x 200) displayed above the page title. This will replace the parent theme/subtheme''s thin banner, if it has one.'
```
johangant, line 14: Is this a double single-quote character a typo or does one of them act as an escape character that when processed results in a single character appearing in the UI?

neilblair: I used the UI to set this Johan.  Assume it is an escape method.  Yes would have thought the escape method would be \' ...

PR #378, opened 2020-07-03
D8NID-662

DuttonMa:
Enable auditing for appropriate content types


DuttonMa:
> I don't think we need to audit the telephone lookup entities

It's OK, we aren't auditing these. It is just confusing the way that this is stored!
Looking at the config file, the content types that are actually being audited are 'contact', 'article', 'health condition' and 'page'.
Note that the ones with a zero next to them are _not_ being audited.

omahm:
> > I don't think we need to audit the telephone lookup entities
> 
> It's OK, we aren't auditing these. It is just confusing the way that this is stored!
> Looking at the config file, the content types that are actually being audited are 'contact', 'article', 'health condition' and 'page'.
> Note that the ones with a zero next to them are _not_ being audited.

Ahh, that is a bit confusing!

PR #379, opened 2020-07-03
Swap over from nidirect_media to origins_media

johangant:


johangant:
NB: PR will fail until origins repo has a new release and this project can reference it. Here as the final step.

Instructions on merge steps are on the original ticket.

PR #380, opened 2020-07-07
D8NID-765 Introduce publications display flag for landing pages

johangant:


johangant:
Sorry, bit of a YAML bomb :)

It's intended to introduce flagged publications into taxonomy term listings and also the related content sidebar.

johangant:
> LG, would be nice to get more descriptive machine names for the flag relationships

Agreed. I can't see a machine name option in the UI to do this. Could be a views UI issue or could be a flag module oversight. They'll work for now and I guess it's possible to match them up to the titles/descriptions with a bit of scrolling. A bit annoying but functional 👍

PR #383, opened 2020-07-08
Add full width image styles, responsive image styles and view modes -…

neilblair:
… these will be the defaults when users embed images in ckeditor before applying the various landscape/portrait image styles. 

What has this got to do with fixing video embed?  When you embed any kind media in ckeditor, it is wrapped in a custom tag "drupal-media".  In the case of an image, by default the media would get the "embed" view mode - but crucially this view mode is not specified as a data attribute in the drupal-media tag when initially embedded.  And this was a problem because I wanted images to float right by default.  So in CKEditor custom CSS,  I made the drupal-media wrapper float right to make it look like landscape_float view mode (which uses the same responsive image style as the embed view mode).  This was silly, because it had a knock-on effect for embedding videos - they would also be wrapped in a drupal-media tag and get floated to the right.

So this PR changes all that so that:

- The default view mode is now article_full for images
- (drupal-media is no longer floated right - separate PR in relation to that - https://github.com/dof-dss/nicsdru_nidirect_theme/pull/136)
- article_full view mode has connected image styles and responsive image styles so that if authors do not apply the usual landscape and portrait view modes, all is not lost - the article_full view mode will output a responsive / optimised image.


johangant:
Cheers @neilblair. I think the top-level config changes here make sense, j'approve! 👍

PR #387, opened 2020-07-17
Fix inconsistency with search API config rendered item field

johangant:


johangant:
Unsure quite why this started happening but a re-check of some very well hidden options and a local re-index seems to solve this.

PR #388, opened 2020-07-20
D8NID-796 migrations on psh

johangant:
I don't think there's any harm in these being enabled by default until we're ready to go live. Open to suggestions, either way!

### DIFF
--- a/migrate-unpack.sh
+++ b/migrate-unpack.sh
@@ -1,4 +1,8 @@
 #!/usr/bin/bash
+if [ ! $MIGRATE_ENABLED == 1 ]; then
+  exit 0
+fi
+
 echo "Restore /app/imports/nidirectd7.sql.gz into drupal7db database..."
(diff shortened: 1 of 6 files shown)


migrate-unpack.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/bash
+echo "Restore /app/imports/nidirectd7.sql.gz into drupal7db database..."
+gunzip /app/imports/nidirectd7.sql.gz
```
johangant, line 3: I don't like this very much, but the more efficient (unpacks into a stream to the mysql client) `gzip -dc /app/imports/nidirectd7.sql.gz | drush -r /app/web sqlc --database=drupal7db` seemed to go dead at some point along the way. Possibly the stream was interrupted by something, it's not clear.

Either way, a manual unpack + restore + tidyup works fine.

PR #389, opened 2020-07-21
D8NID-800 scheduled dev build

johangant:
This PR does the following:

- Every night, Mon - Fri, it kicks off an 'edge build' that takes the development branch, swaps all dof-dss packages to use HEAD on `development` branches, then pushes the contents into a branch called  `D8NID-edge`. This is then carried over to platform.sh via the webhook integration in use at the moment where a build/deploy task is triggered. The environment is active and won't normally be pulled down until it's no longer needed, eg: site being live.
- Half an hour later, a second job runs a platform sync task on that environment to update the content from the master environment.

Circle CI doesn't push a throwaway branch to the platform.sh repo, because when I originally did this the github integration seemed to prune my branch some way through the build process and got in the way. As a result, if Circle CI pushes back to GitHub the edge branch ends up following the route of other code branches.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -340,11 +340,11 @@ workflows:
     triggers:
       - schedule:
           # At 00:00 on every day-of-week from Monday through Friday
-          cron: "0 12 * * 1-5"
+          cron: "0 0 * * 1-5"
           filters:
             branches:
               only:
-                - D8NID-800-nightly-build
+                - development
     jobs:
       - build
       - dev_build:
@@ -357,11 +357,11 @@ workflows:
   nightly-content-sync:
     triggers:
       - schedule:
-          # At 00:10 on every day-of-week from Monday through Friday
-          cron: "5 12 * * 1-5"
+          # At 00:30 on every day-of-week from Monday through Friday
+          cron: "30 0 * * 1-5"
           filters:
             branches:
               only:
-                - D8NID-800-nightly-build
+                - development
     jobs:
       - sync_data
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -340,7 +340,7 @@ workflows:
     triggers:
       - schedule:
           # At 00:00 on every day-of-week from Monday through Friday
-          cron: "0 0 * * 1-5"
+          cron: "30 11 * * 1-5"
           filters:
             branches:
               only:
@@ -365,10 +365,10 @@ workflows:
     triggers:
       - schedule:
           # At 00:10 on every day-of-week from Monday through Friday
-          cron: "10 0 * * 1-5"
+          cron: "35 11 * * 1-5"
           filters:
             branches:
               only:
-                - D8NID-edge
+                - D8NID-800-nightly-build
     jobs:
       - sync_data
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,12 +9,12 @@ commands:
           name: Add GitHub access token for composer
           command: echo 'composer config -g github-oauth.github.com $GITHUB_TOKEN' >> $BASH_ENV
       # Add SSH user key so we can access related repositories as part of our initial clone + composer install command.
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -9,12 +9,12 @@ commands:
           name: Add GitHub access token for composer
           command: echo 'composer config -g github-oauth.github.com $GITHUB_TOKEN' >> $BASH_ENV
       # Add SSH user key so we can access related repositories as part of our initial clone + composer install command.
-      # 89:cc >> GitHub user key fingerprint
-      # ed:e9 >> private fingerprint to talk to platform.sh
+      # f8:f0 >> private fingerprint to allow Circle CI to talk to platform.sh.
```
johangant, line 12: Swapped the order as otherwise commits would end up looking like they came from me. This way, Circle CI's key takes priority.

.circleci/config.yml
```diff
@@ -240,120 +240,91 @@ jobs:
...
+              dof-dss/nicsdru_origins_modules:dev-development dof-dss/nicsdru_nidirect_theme:dev-development
+
+            git add composer.*
+            git commit -m "Set dof-dss packages to HEAD development for build"
+
+            # Push to fixed, non-integrating build branch. GitHub webhook integration will propagate this
+            # to platform.sh for later steps to use.
+            git push -f origin $DEV_BUILD_BRANCH
+
+            # Pause for webhook to propagate the change from GH to PSH.
```
johangant, line 288: Might not be needed, was there originally to act as a pause before trying to sync data from master. Given that any CI checks can take 10 mins and the deploy task takes around 2 minutes, this is why the sync task has been decoupled to its own workflow, timed to run at a later point.

.circleci/config.yml
```diff
@@ -240,120 +240,91 @@ jobs:
...
-      - github_hosts_workaround
       - checkout_code
       - install_php_os_extensions
-      - composer_tasks__no_cache
       - run:
-          name: PHPCS static analysis
-          command: /home/circleci/project/phpcs.sh /home/circleci/project "/home/circleci/project/web/modules/origins /home/circleci/project/web/modules/custom /home/circleci/project/web/modules/migrate /home/circleci/project/web/themes/custom"
+          name: Add extra OS and PHP extensions/config
+          command: |
+            sudo docker-php-ext-install pcntl posix
```
johangant, line 308: Needed by the platform CLI tool

.circleci/config.yml
```diff
@@ -240,120 +240,91 @@ jobs:
...
+            mkdir -p ~/.ssh
+            ssh-keyscan -H ssh.$PLATFORM_REGION >> ~/.ssh/known_hosts
+      - run:
+          name: Install the Platform.sh CLI tool
+          command: |
+            curl -sS https://platform.sh/cli/installer | php
+      - run:
+          name: Trigger a data sync from master environment to nightly edge build.
+          command: |
+            /home/circleci/.platformsh/bin/platform sync data -p $PLATFORM_PROJECT_ID -e $DEV_BUILD_BRANCH -y
```
johangant, line 321: Things like $DEV_BUILD_BRANCH are stored in Circle CI's environment variable store. You can't see the value once created, so to update you basically have to remove + recreate. I think you can see the value if you run a job with SSH enabled, then SSH into the running container and use the `env | sort` trick to show you the values stored in the OS.

.circleci/config.yml
```diff
@@ -381,6 +347,21 @@ workflows:
                 - development
     jobs:
       - build
-      - functional_js_tests__config_only:
+      - dev_build:
```
johangant, line 350: D'oh, this didn't quite catch my marketing skills. I started with a 'dev' build but as that's quite ambiguous, I preferred to stick with 'edge build'. Probably not worth changing here unless it really bugs or confuses someone.

johangant: It bugged me so I fixed it :)

PR #392, opened 2020-07-24
Add placeholder region to views header

johangant:
Allows us to add a convenience link in nidirect_news.module

Relates to https://github.com/dof-dss/nidirect-site-modules/pull/199


johangant:
Think I'm going to rework this so it's added in preprocess rather than needing this empty placeholder here.

johangant:
Performing a u-turn here. Preprocessing the underlying node doesn't quite work as we render 4 news items, not the featured content list node itself. And programmatically adding a footer + contents is actually rather fiddly so this is a handy practical solution.

PR #396, opened 2020-07-30
D8 nid 788

omahm:


omahm:
> LGTM. A later PR needed to remove taxonomy_manager from composer once it has uninstalled cleanly?

I'm one step ahead, created https://digitaldevelopment.atlassian.net/browse/D8NID-813

omahm:
Not going to merge this until we have removed the table preprocess code from origins_taxonomy_access and removed origins_taxonomy_access as it is now redundant but there is a dependency on taxonomy_access_manage which prevents us from uninstall that module.

PR #404, opened 2020-08-18
Fix repository version update notice; allows mariadb-client to cleanly install

johangant:


johangant:
Pushing this one through as it's blocking other PRs

PR #408, opened 2020-08-21
D8NID-848 Add block_content_permissions module + related config

johangant:
Without this, editors need to be granted the 'administer blocks' permission which is not appropriate. There is no specific route or access controller to override/hook into so this contrib module does a great job at isolating the custom block types that specific roles can create and bridges the gap we have, with minimum fuss.


DuttonMa (commented):
Looks good, but are you sure that you only need to change permissions for editors and admins ? Aren't there lots of other roles that should be considered ?

johangant:
> Looks good, but are you sure that you only need to change permissions for editors and admins ? Aren't there lots of other roles that should be considered ?

Currently following the wider LB patterns which are set to both. I'm sure they'll be expanded and refined as time goes on but that's the only direction I've got at the moment, and it looks sensible as a starting point at least.

PR #415, opened 2020-09-01
D8NID-844 Order homepage site themes based on taxonomy weight

johangant:
Taxonomy navigator module uses the weight property, rather than a discrete field value.


DuttonMa (approved):
Looks good

PR #418, opened 2020-09-02
D8 nid 830

omahm:


johangant (approved):
The CI task will fail here because the module's not present in the version referenced by composer. It should work with the nightly build though or until our next version bump on nidirect-drupal.

omahm:
I also updated Drupal console from part of the work around exploring content export

DuttonMa (approved):

PR #420, opened 2020-09-04
Forces a full purge of environment solr contents before reindexing.

johangant:
Shouldn't really be needed but this ensures the index is completely
empty to avoid issues of double-counting seen several times even after
syncing from master environment.


omahm (approved):
Hopefully this will fix the Solr count issues!

PR #423, opened 2020-09-07
Change field_theme_summary to long, plaintext field

johangant:
Requires migration rollback before coming into effect or config import will fail


omahm (approved):
👍

PR #424, opened 2020-09-08
Split solr purge and rebuild commands to avoid input prompt

johangant:


johangant:
Trying to see if I can run this via the Circle CI API rather than wait for an overnight workflow trigger to happen

omahm (approved):
Nice idea

PR #425, opened 2020-09-08
Circle ci fix solr rebuild

johangant:
Missed the obvious on the first command: it was missing the project and environment parameters so the later commands were never reached.


omahm (approved):
👍 that'll work

PR #426, opened 2020-09-08
Switch the migrate asset process from Azure to PSH + Circle CI

johangant:
We use Circle CI to define a weekly trigger workflow (Saturday) and then use the PSH CLI tool to fetch databases and files. The database is unpacked into the legacy db service using drush, executed with the CLI tool. Managed files are retrieved with the CLI tool's mount command which uses rsync behind the scenes to ensure we can avoid fetching the entire 5GB or so of files each time and just fetch any changes.

Executes with the NICS user's CLI token, as per the other commands in other workflows.


omahm (approved):
LG 👍

johangant:
Needs a change as this will pull assets onto Circle CI, then sync. Instead I think I want to use Circle CI to drive it but rely on scripts/tools running on the PSH cluster

PR #440, opened 2020-09-16
Change to better container, add platform CLI tool, run against existing instances

johangant:
The container image needed to be changed so we can drop in our custom PHP extensions/config. It also needed the platform CLI tool installing.

The missing link is the CLI token env var, added to the D8 application master environment that allows the NICS PSH user to access services between the D7 and D8 projects.


omahm (approved):
LG

PR #444, opened 2020-09-21
D8 nid 857

omahm:


johangant (approved):
Spotted this online not long after the original mention of it. Seemed like a good choice at first glance so I'm pleased we've not re-invented the wheel :)

omahm:
@neilblair I've added some user permissions based off the description in D8NID-333, can you review the role config in this PR  to see if we have those covered?

PR #445, opened 2020-09-21
D8NID-687

DuttonMa:
Install Fastly module on production only.
Also set Fastly site id to 'master' (rather than the randomly generated id which fails validation).


omahm (approved):
LG

johangant (approved):
👍

PR #448, opened 2020-09-22
Update drupal-check + phpstan libraries

johangant:
It keeps us in line with the latest releases but it does report errors autoloading test classes for some reason. You can work around this with a `-e "*/tests/src/*"` which isn't great but it means we're still able to detect deprecated code in our non-test classes until this quirk is fixed by us or upstream


omahm (approved):

DuttonMa (approved):
Looks good

PR #451, opened 2020-09-24
D8NID-873

DuttonMa:
Remove cache tags from embedded search view display.
Enabling cache tags on this display was causing the 'search_api_list:default_content' cache tag to be added to every page on the site. This cache tag is invalidated when any node is created, updated or deleted so this was not desired behaviour.
Investigation has revealed that adding the search box to a page does not necessitate adding any extra cache tags.


johangant (approved):
Good work! I wonder if this is a common issue on Drupal 8 sites or is this just the way we happen to have arrived at our present setup in NIDirect?

PR #452, opened 2020-09-24
D8NID-154 health conditions a-z

neilblair:
Re-order blocks so that facet drop-downs appear above A to Z.
Change name of body location facet to body part.


config/sync/views.view.health_conditions.yml
```diff
@@ -295,6 +297,10 @@ display:
       display_extenders: {  }
       path: services/health-conditions-a-z
       exposed_block: true
+      cache:
+        type: none
```
omahm, line 301: Shouldn't we have caching by parameter/tag?

omahm (commented):
Looks good but I think we should be caching

config/sync/views.view.health_conditions.yml
neilblair: Yeah not sure why this is set to none - wasn't something I altered in the view.  Will take a look

johangant (commented):
Looks fine to me, thanks. +1 for the addition of views caching with search API tags.

omahm (approved):
LGTM 👍

PR #455, opened 2020-09-25
D8 nid 354

DuttonMa:
Changes to GP admin views after QA failure:
- change pager
- change 'surname' to 'last name'
- add extra columns to enable sorting by 'online appointments' and 'online prescriptions' (extra code required to make this work in PR https://github.com/dof-dss/nidirect-site-modules/pull/245


config/sync/views.view.gp_practice_list.yml
```diff
@@ -2053,6 +2069,768 @@ display:
...
+          admin_label: ''
+          order: DESC
+          exposed: false
+          expose:
+            label: ''
+          entity_type: node
+          entity_field: title
+          plugin_id: standard
     cache_metadata:
       max-age: -1
```
johangant, line 2835: Does this mean once generated the views output never expires unless forcibly flushed? 

That's probably fine if so... better to render once and cache, then only regenerate if a GP or related thing changes.

johangant (approved):

config/sync/views.view.gp_practice_list.yml
DuttonMa: I'm not entirely sure what this means tbh, but the caching for both views is set to 'tag based' in the normal way so I'm happy with that.

PR #462, opened 2020-10-05
D8NID-855 ~ Remove banner field and overlay image from landing page

omahm:


johangant (approved):
We're not migrating these anymore, are we? If not, might be wise to mark the corresponding migrations as disabled or remove them altogether?

omahm:
> We're not migrating these anymore, are we? If not, might be wise to mark the corresponding migrations as disabled or remove them altogether?

Good point, I'll update mig mods

PR #465, opened 2020-10-07
D8NID-842

DuttonMa:
change cache max age to 12 hours on production


johangant:
Would it be wise to increase this to the absolute maximum? A fairly generic D8 pattern is to do this, then rely exclusively on cache tags and other aspects of the cache API to only expire items that change or are purged with intent, rather than by TTL alone.

DuttonMa:
> Would it be wise to increase this to the absolute maximum? A fairly generic D8 pattern is to do this, then rely exclusively on cache tags and other aspects of the cache API to only expire items that change or are purged with intent, rather than by TTL alone.

Yes, it is a fair point. I was just making the changes as specified in D8NID-842 - we can raise it in the standup this morning.

johangant (approved):
I'll approve in advance: we discussed + agreed aiming high for cache lifetimes, with a view to debug + fix in instances where something isn't quite right.

DuttonMa:
Page cache times now set to the maximum of one day as agreed.

DuttonMa:
After some experimentation and discussion, have now found that it is possible to set whatever max cache time you like in config and Drupal respects this.
Have now set the max cache time to one year.
In this case the admin/config/development/performance admin screen was misleading as it showed 'no cache' even though 1 year had actually been selected. Have created another PR https://github.com/dof-dss/nidirect-site-modules/pull/250 which alters the performance form accordingly to avoid confusion.

PR #466, opened 2020-10-07
D8NID-900 migrate config sync and full import

johangant:
This PR aims to:

- Provide a script we can use to sync the db config from the migrate modules (excluded from the main config/sync folder due to past config import issues).
- Full rollback + re-import of content. Migrate import alone is not enough to 'top-up' the content as there are deletions to contend with. It's far safer (although time consuming) to roll it all back, re-import and re-add any extras we need on top of that.

### DIFF
--- a/migrate-full-import.sh
+++ b/migrate-full-import.sh
@@ -1,4 +1,16 @@
 #!/bin/bash
+
+# ABOUT
+# This is a selective rollback + import script that assumes a starting point of all generic
+# D7 upgrade migrations having been done/refined, then custom migrations already imported at
+# least once before. This script rolls back all the key migrations, in the order required,
+#
+# Migrate top-ups aren't quite so useful for us here, because source content removals
+# are not also kept in sync with the D8 site, so a full rollback + re-import is safest for
+# data integrity.
+#
+# This process can take a LONG time to run (hours) so please be patient.
+
 if [ $LANDO="ON" ]; then
   DRUPAL_ROOT=/app/drupal8/web
 else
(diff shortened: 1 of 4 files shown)


omahm (approved):
Nice one, great job 👍

migrate-full-import.sh
```diff
@@ -0,0 +1,79 @@
...
+# Import files.
+drush migrate:import upgrade_d7_file --execute-dependencies
+# Import media entities
+drush migrate:import --group=migrate_drupal_7_file
+
+# Import GP entities.
+drush migrate:import --group=migrate_nidirect_entity_gp
+
+# Import all node migrations.
+for type in driving_instructor application article external_link gp_practice health_condition landing_page news nidirect_contact contact page publication; do
```
DuttonMa, line 66: Think you need to add the new 'webform' node migration into this list

migrate-full-import.sh
```diff
@@ -0,0 +1,79 @@
...
+##### ROLLBACK ######
+
+# Import URL aliases and redirects
+drush migrate:rollback --group=migrate_drupal_7_link
+
+# Rollback book migration.
+drush migrate:rollback nidirect_book
+
+# Rollback all node migrations.
+for type in driving_instructor application article external_link gp_practice health_condition landing_page news nidirect_contact contact page publication; do
```
DuttonMa, line 35: Think you need to add the new 'webform' node migration into this list

DuttonMa (changes requested):
Really nice work, just a couple of omissions I think

migrate-full-import.sh
johangant: Yes, good reminder. I'll add that

johangant: Done!

PR #473, opened 2020-10-09
D8 nid 896

omahm:


DuttonMa:
Although this fix will rectify the problem in the bug report (for an editor user), an unwanted side effect is that if you log in as an author you will also now see the 'Change to Published' link which is not correct as an author does not have sufficient privileges.

The view is actually working as designed.
The custom text field 'nothing' contains links to 'Change to Draft' and 'Change to Published'. 
The custom text field 'nothing_3' contains just 'Change to Draft'.
Only one of the custom text fields 'nothing' and 'nothing_3' should ever be shown, they are mutually exclusive.
The choice of which one to show is made according to the current user's permission level.

The logic may be found in the origins_workflow module, in the function origins_workflow_hide_moderation_field(). This is where the problem lies. A quick look at this function reveals that this line:

    elseif ($view->element['#display_id'] == 'needs_review') {

will never be reached as this 'if' statement at the top of the function:

   if (!empty($view->element['#name']) && $view->element['#name'] == 'workflow_moderation') {

.. will always be true.

So it seems that the solution to the bug here is to move the 'elseif' inside the 'if'.
The logic will then be 'if view name is 'workflow_moderation' then check display id. If display is 'my drafts' or 'all drafts' do this, if display is 'needs review' do that etc.

DuttonMa (changes requested):
Changes needed as described above

DuttonMa (approved):
Looks good now

PR #477, opened 2020-10-16
Health conditions search tuning:

neilblair:
- Change parse mode for the search terms to "multiple terms" rather than "phrase" in the view
- Enable partial word matching on many fields
- Alter boosts for fields similar to D7 version
- Remove html filter boosting

These changes should produce a fairly similar search results to the present D7 version.


omahm (approved):
LG

PR #478, opened 2020-10-19
Remove schema metatags for medical conditions

johangant:
Patch isn't fully stable with 2.1 release, causing fatal exceptions in other parts of the build

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -176,8 +176,7 @@
                 "SHS select is empty on views": "https://www.drupal.org/files/issues/2018-12-10/shs-blank-select-fix-3019378-3.patch"
             },
             "drupal/schema_metatag": {
-                "Add @id as a tag to all schemas": "https://www.drupal.org/files/issues/2019-09-01/3078582-02.patch",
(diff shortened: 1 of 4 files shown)


DuttonMa:
So what do we need to do here? Do we need to create another patch to replace the one that is being removed?

johangant:
> So what do we need to do here? Do we need to create another patch to replace the one that is being removed?

That's the idea, but that task has been split into its own ticket as it's fairly self-contained. The wider schema_metatags functionality is useful to keep in.

omahm:
> > So what do we need to do here? Do we need to create another patch to replace the one that is being removed?
> 
> That's the idea, but that task has been split into its own ticket as it's fairly self-contained. The wider schema_metatags functionality is useful to keep in.

Might also need to look at how this affects taxonomy terms.

johangant:
> Might also need to look at how this affects taxonomy terms.

Reviewed against the original bug reported and can confirm this solves that issue. There shouldn't be any extra metatags set by this module for terms but I think it tickled the bug when it tried to load defaults with the entity (or related hooks) when requesting the taxonomy term edit form. Either way, this looks like a safe merge until the medical condition patch can be re-rolled.

omahm (approved):
👍

PR #479, opened 2020-10-19
D8NID-822 sidebar featured news block

johangant:
A refactored approach to allow us to display featured news in the sidebar. A straightforward block display was scuppered because you cannot sort a view's results by the order of contextual arguments and jiggling about the query with a select sub-query and/or preprocessing the results was not going anywhere.

So instead we effectively recycle the code already written for the views landing page to present the order of nodes defined in the FCL node at /news. A slight adjustment was needed to specify the active display and the display uses an embed to let us borrow as much front-end markup and styling as possible.

With this approach, we can now consistently use the FCL node to define the order of news items on the landing page and any sidebar content.


DuttonMa (approved):
Pretty hard for a human to validate that Yaml, but the explanation makes sense !

PR #480, opened 2020-10-20
D8 nid 915

DuttonMa:
Had a look through all views to find those using the 'node_list' cache tag and replaced it with a 'handy cache tag' where possible (in order to greatly improve performance).
Views changed:
- Contacts A-Z
- Health conditions A-Z
- News views


johangant (approved):
Nice one, makes sense to add those here.

PR #481, opened 2020-10-23
D8NID-912 Update Solr field analyzer config

johangant:
- Improves searching over text fields; has token cleanup, synonym support and stemming in place. Adapted legacy Solr config to use components that Solr 7 provides.


johangant:
Holding off merging this until we're sure the summary field updates/switcheroo is okay

PR #482, opened 2020-10-26
DNID-436 metatags refinements

johangant:
- Removes driving instructor config (too little content to be of reasonable use)
- Adds extra config for News including previously missing photo field


omahm (approved):
LG

PR #486, opened 2020-10-29
D8NID-921 ~ Elevated Solr ID config

omahm:


johangant:
Great stuff, I think I can see how this works. What picks up the YML files, is it how search API already handles things?

omahm:
> Great stuff, I think I can see how this works. What picks up the YML files, is it how search API already handles things?

They're config entities, this is simply importing them.

johangant (approved):
Good stuff, let's get it in the build 👍

DuttonMa (approved):
Looks good

PR #495, opened 2020-11-09
D8 nid 916

omahm:


johangant (approved):
LGTM. Shouldn't pose any issues for the contacts search either

PR #503, opened 2020-11-13
Don't show content flagged as 'Hidden'

johangant:
Flag module ids leave something to be desired... ambiguous++. 

Not all bad though, as the admin labels are very clear and anyone working with the views UI shouldn't struggle with that.


omahm (approved):
LG

PR #506, opened 2020-11-16
Adds regex validation pattern to email field.

neilblair:
The change is buried in line 15 of the yml file!

`enter_a_valid_email_address:\n  '#type': email\n  '#title': 'Enter a valid email address'\n  '#required': true\n  '#required_error': 'You must enter an email address so that we can reply to you.'\n  '#pattern': '[^@\\s]+@[^@\\s]+\\.[^@\\s]+'\n  '#pattern_error': 'You must enter a valid email address, for example, name@example.com'`


johangant:
Looks good. NB: this looks like massive changes in the file but it's due to the presence, or lack of presence, of the `multiline_config` which would apply if local development config is active. It's really not too much of a problem, as the end result is the same, but it can make for a confusing or hard to read PR if you, say, dump from production and immediately import to Lando without also updating active config.

johangant (approved):

neilblair:
> Looks good. NB: this looks like massive changes in the file but it's due to the presence, or lack of presence, of the `multiline_config` which would apply if local development config is active. It's really not too much of a problem, as the end result is the same, but it can make for a confusing or hard to read PR if you, say, dump from production and immediately import to Lando without also updating active config.

I think I'll need to ask you about that tomorrow! I think there may be a few oddities with my local setup.

PR #510, opened 2020-11-19
Update core to 8.8.11 to fix https://www.drupal.org/sa-core-2020-012

johangant:
Also shift patch for file transliteration due to stagnation in the original drupal.org issue. 

Patch re-rolls for https://www.drupal.org/project/drupal/issues/2492171 now moved to https://www.drupal.org/project/drupal/issues/3094052

Version bump for search_api_spellcheck as it now contains the patch we previously used with the alpha release, which wouldn't then apply cleanly.


DuttonMa (approved):
Looks good

PR #517, opened 2020-11-24
Remove scheduler fields from form

johangant:
This node type isn't subject to content moderation as it's a supporting node type.

This change also cleans up the inline entity form when adding these nodes to existing HC nodes


johangant:
Superseded by changes and fixes in https://digitaldevelopment.atlassian.net/browse/D8NID-981

PR #533, opened 2020-12-03
D8NID-785

johangant:
Adjusts contact search from the D7 implementation so it only operates on the contact title. This is useful because geographical terms like 'Belfast' often return content titles that don't look relevant, purely because they have address content in the body field that match the search term rather than a more contextually relevant result around what that org/contact does.


neilblair (approved):
LG

PR #534, opened 2020-12-03
Adds patch for telephone_plus 

neilblair:
Patch adds phone extension into link text of phone links.

https://www.drupal.org/project/telephone_plus/issues/3186246

https://www.drupal.org/files/issues/2020-12-03/3186246-1.patch


johangant (approved):
Needs the lock file updating, but otherwise all good!

PR #538, opened 2020-12-08
D8NID-964: Adjust HC author/supervisor permissions and IEF settings

johangant:


omahm (approved):
Hiding the form fields, that makes sense!

PR #541, opened 2020-12-14
D8 nid 1013

omahm:


config/sync/core.extension.yml
```diff
@@ -28,6 +28,7 @@ module:
   datetime: 0
   datetime_range: 0
   diff: 0
+  dynamic_entity_reference: 0
```
johangant, line 31: Didn't expect to see this module for scheduled transitions. Is it a dependency?

johangant (approved):
Looks good to me, although had a question about the other enabled module in the core.extension.yml file.

config/sync/core.extension.yml
omahm: Yup!

PR #551, opened 2020-12-22
D8NID-996: minor adjustments

johangant:
* Add caption field to form so you can populate it when adding a new image


DuttonMa (approved):
Looks good

PR #558, opened 2021-01-08
Enables linkit on link fields in various entity form displays.

neilblair:


config/sync/core.entity_form_display.node.news.default.yml
```diff
@@ -72,11 +72,12 @@ content:
     settings:
       placeholder_url: ''
       placeholder_title: ''
+      linkit_profile: default
     third_party_settings: {  }
-    type: link_default
+    type: linkit
     region: content
   field_summary:
-    type: text_textarea
```
neilblair, line 79: Not sure why field_summary is being changed from text_textarea to string_textarea here - but I think string_textarea is correct

config/sync/core.entity_form_display.node.news.default.yml
```diff
@@ -72,11 +72,12 @@ content:
...
+    type: string_textarea
```
neilblair, line 80: Not sure why field_summary is being changed from text_textarea to string_textarea here - but I think string_textarea is correct

johangant: Not sure either, but if it's autogenerated by config export it's entirely possible that's a change from core itself.

johangant (approved):
LGTM. LinkIt taking over :)

PR #560, opened 2021-01-12
Swap to full pager, add taxo desc field

johangant:
Full pager fixes broken total result count, desc field from taxonomy term entity allows searching by keyword such as Belfast, rather than BT1 alone


omahm (approved):
LG

PR #598, opened 2021-02-05
Drop internal cron so we can use external service

johangant:
We're likely going to use GCP Scheduler to kick off the HTTP requests for cron as it can reliably execute every 60 seconds.


omahm:
Is the GCP cron set to fire against all of our PSH instances?

DuttonMa (approved):

johangant:
> Is the GCP cron set to fire against all of our PSH instances?

@neilblair has just configured it this morning

PR #599, opened 2021-02-05
D8NID-1074

DuttonMa:
Apply patch to scheduled transitions to make revisions in any language available for scheduling


johangant (approved):
👍  just needs the lockfile updated before merging

DuttonMa:
> 👍 just needs the lockfile updated before merging

Oops, yes, forgot about that ! Doing that now, thanks

PR #606, opened 2021-02-12
Add file mime relationship and field to media list display

johangant:
Preprocessed in origins_media.module to make it prettier

After adding the extra column, looks like:

Screenshot 2021-02-12 at 12 01 54


omahm (approved):
LG

PR #612, opened 2021-02-19
D8 nid 1093

omahm:
Multiline config strikes again.


config/sync/core.extension.yml
```diff
@@ -187,7 +188,6 @@ module:
   twig_field_value: 0
   twig_tweak: 0
   ultimate_cron: 0
-  update: 0
```
johangant, line 190: Looks like this is tripping config install errors. Might be worth leaving in here and removing later in another ticket, if needed.

johangant (approved):
👍  I added multiline_config into the standard extensions and pulled them out of the local config split profile. Should make it easier to move changes between environments now

PR #621, opened 2021-02-26
D8NID-1102 upgrade to PHP 7.4

johangant:
Moves us closer towards a smoother Drupal 9 adoption too by:

* PHP 7.4 for longer coverage of security releases and improved performance
* Adds composer 2 for faster builds
* Adds Redis 6 with latest PhpRedis (5.3.3)
* Optional verbose error output on platform.sh environments
* Latest platform.sh config (compatible with Drupal 9)

PhpRedis tags can be seen at https://github.com/phpredis/phpredis/tags

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -9,14 +9,14 @@ name: 'nidirect'
 # The runtime the application uses.
 type: 'php:7.4'
 
-dependencies:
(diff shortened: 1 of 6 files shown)


omahm (approved):
Great stuff!

PR #625, opened 2021-03-03
D8NID-1108 ~ Added Taxonomy Entity Index rebuild task

omahm:


johangant (approved):
Looks good. That should work because the drush command doesn't take any arguments, which previously needed it to be wrapped in double quotes.

:shipit:

PR #629, opened 2021-03-05
D8NID-1105

johangant:
Moving the links for moderation state change from views config markup to links in the operations column. See https://github.com/dof-dss/nicsdru_origins_modules/pull/133 for details.


johangant:
Forgot to add reviewers on Saturday, doh!

omahm (approved):

composer.json
```diff
@@ -30,7 +30,7 @@
         "composer/installers": "^1.9",
         "cweagans/composer-patches": "^1.6.5",
         "dof-dss/nicsdru_nidirect_theme": "^0.12.1",
-        "dof-dss/nicsdru_origins_modules": "^0.9.13",
+        "dof-dss/nicsdru_origins_modules": "dev-D8NID-1105",
```
DuttonMa, line 33: I assume that this is a mistake and that you meant to change it back ?

DuttonMa (changes requested):

composer.json
johangant: Yep, exactly right. I need this in place right now otherwise the project builds using the tag which means I can't combine it with the corresponding branch on the other repo.

Last time I forgot to update before merging so making an extra note to do it on this PR to save a hasty commit to follow it :)

PR #634, opened 2021-03-08
Use 9.x compatible version of the module, and patch to fix drush command error

johangant:
An expected module update, but need the patch in there too to fix an otherwise fatal error when enabling some modules like webprofiler.


johangant:
Hold the fort... looks like a broken service reference too. I'll roll a new release for 2.x and reference that

PR #636, opened 2021-03-10
D8NID-1121 Add fpa module, turn off dblog, amend config profiles

johangant:
- Drop dblog from local config (causes complications when switching between config split profiles)
- Set fpa module to run on local split


omahm (approved):
LG

PR #638, opened 2021-03-12
D8NID-1111 minor performance tweaks and changes

johangant:
Based on findings from Blackfire profiling.


web/sites/default/settings.php
```diff
@@ -71,6 +71,12 @@
       $settings['container_yamls'][] = $app_root . '/' . $site_path . '/services.yml';
       break;
 
+    case 'D8NID-edge':
+      // Retain as much prodution related config/settings as possible.
```
DuttonMa, line 75: spelling mistake here 'prodution' - very picky and tiny I know !

DuttonMa (approved):
Good stuff, great to get this sorted !

web/sites/default/settings.php
johangant: Good spot, thanks!

PR #640, opened 2021-03-16
D8NID-1129

johangant:
NB: COVID-19 block node id liable to shift with migrations, so block config may need to be updated to reflect a shifting high water line for content ids after each migration run.


johangant:
Sigh... TFW when you realise your git branch references the wrong ticket id right at the very end 🤦

PR #641, opened 2021-03-16
D8NID-1127

johangant:
NB: COVID-19 block node id liable to shift with migrations, so block config may need to be updated to reflect a shifting high water line for content ids after each migration run.

### DIFF
--- a/config/sync/block.block.featuredcontent_covid19.yml
+++ b/config/sync/block.block.featuredcontent_covid19.yml
@@ -1,4 +1,4 @@
-uuid: 52363a5e-401c-4107-99d1-454125821310
+uuid: 6d3d63bc-7093-47ef-a2e8-8f826c33339f
 langcode: en
 status: true
 dependencies:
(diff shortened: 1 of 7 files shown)


config/sync/block.block.featuredcontent.yml
```diff
@@ -18,6 +18,7 @@ settings:
   label: Featured
   provider: nidirect_homepage
   label_display: visible
+  featured_items: '14474'
```
DuttonMa, line 21: Could we automate the setting of this? Feels a bit risky to rely on us remembering to set it manually after each run (or have I misunderstood ?)

DuttonMa (changes requested):

config/sync/block.block.featuredcontent.yml
johangant: Handled with env vars in https://github.com/dof-dss/nidirect-site-modules/pull/353

PR #643, opened 2021-03-18
Translation help link

neilblair:


web/sites/example.settings.local.php
```diff
@@ -131,15 +131,15 @@
 $settings['skip_permissions_hardening'] = TRUE;
 
 /**
- * Exclude modules from configuration synchronisation.
+ * Exclude modules from configuration synchronization.
```
DuttonMa, line 134: Why change this from English to American ?

DuttonMa (approved):
All looks good to me

web/sites/example.settings.local.php
neilblair: I wondered that myself!

PR #652, opened 2021-03-25
D8NID-1140 allow editors to reference unpublished taxonony term entities

johangant:
Fine for admins who have the bypass access checks permission. Not so for other roles, so this plugin helps as an interim solution until core sorts itself out and makes this configurable :)


neilblair (approved):
👍🏻

PR #655, opened 2021-03-26
D8NID-1141

DuttonMa:
Apply a redirect module patch to add 'URL redirects' tab to taxonomy term edit form


johangant (approved):
Looks good, just needs the lock file updated

PR #657, opened 2021-03-31
Drop upgrade_status module usage; fall back to using phpstan/drupal-check in CI

johangant:
CI steps might fail as I've twiddled the job names and that won't match up with broader repo config. I've done this to split the coding standards from deprecated code checks; these can run in parallel.


DuttonMa (approved):
Good stuff

PR #662, opened 2021-04-01
Landing pages config:

neilblair:
- new contact card block type
- add strapline (field_teaser) to banner deep block
- re-config allowed blocks to remove a few new types that crept in


johangant (approved):
LGTM

PR #667, opened 2021-04-09
D8NID-1152 Revoke some key permissions from auth'd user role

johangant:
Ensures that users potentially misconfigured to only have authenticated user role cannot see the admin theme nor view unpublished content: these are delegated to more specific roles


omahm (approved):
LG

PR #674, opened 2021-04-19
Patch paragraphs to allow editors to edit layout builder paragraph revisions

johangant:
Also peg inline_entity_form to rc8 (for now) to allow its patch to apply cleanly


neilblair (approved):
LGTM

PR #686, opened 2021-05-04
Adjust contacts view filters

johangant:
Split query group into two: one for node title, one for supplementary title.

Dovetails with query alter and view preprocess code in nidirect_contacts.module

Relates to https://github.com/dof-dss/nidirect-site-modules/pull/375


omahm (approved):
LG

PR #695, opened 2021-05-11
D8NID-1189 CSP module and config

johangant:
* Adds the content security policy module with report only config.
* Adds filelog module to avoid spraying a firehose of events at dblog (platform.sh can't support syslog)
* Cron handler to prune away compressed log files over 30 days old.

### DIFF
--- a/config/sync/csp.settings.yml
+++ b/config/sync/csp.settings.yml
@@ -1,0 +1,59 @@
+_core:
+  default_config_hash: o4-eDkmTmVdUICtL1VxWD3S9Ze8kNq_lHj6DskF8U8o
+report-only:
+  enable: false
+  directives:
+    default-src:
+      base: self
+    script-src:
+      base: self
+    script-src-attr:
+      base: self
+    script-src-elem:
+      base: self
+    style-src:
+      base: self
+    style-src-attr:
+      base: self
+    style-src-elem:
+      base: self
+    frame-ancestors:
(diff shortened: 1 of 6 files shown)


neilblair (changes requested):
Looks good Johan - only concern is frame-ancestors settings.  I think we'll need that set to 
`frame-ancestors: 'self' www.youtube.com` or something like that?  Unless that is oembed works ok without that?

config/sync/csp.settings.yml
```diff
@@ -0,0 +1,72 @@
...
+        - www.googletagmanager.com
+      base: self
+    style-src:
+      base: self
+    style-src-attr:
+      flags:
+        - unsafe-inline
+      base: self
+    frame-ancestors:
+      base: any
```
neilblair, line 69: Think frame-ancestors should be self - and might need to specify youtube domain(s) as sources

neilblair (changes requested):

DuttonMa (approved):
Interesting stuff !

config/sync/csp.settings.yml
johangant: Yes, thanks for spotting. I had taken it off earlier to test something and forgotten to reset it. Sorted now

johangant:
Addressed earlier issue raised by @neilblair - merging so it's ready for Monday

PR #696, opened 2021-05-12
Landing pages block updates 2

neilblair:
- Add information notice block for landing pages (to be used instead of ckeditor info-notice style on the text block)
- Re-enable layout builder block restrictions by region to ensure editors cannot, for example, add cards into the one column layout section or add information notice blocks into x2 and x3 card sections.
- Remove p.info-notice style from layout builder text format


johangant (approved):
Looks good, nice one

PR #698, opened 2021-05-18
D8NID-1205 ~ Adding Twig VarDumper to make it easier to debug templates

omahm:
Adding this module as it's been a PITA to debug twig now that Kint isn't a sub-module to Devel. I ran into the old issue that any {{ kint() }} inserted into a template would result in the site redirecting to install.php. Prior to the Devel update you could get around this by requiring the module in settings.php and then calling Kint::$maxLevels = 3; but this doesn't work anymore.


johangant (approved):
Looks useful! 👍

PR #705, opened 2021-05-21
Updated general webform config

johangant:
Filled in the gaps based on vanilla webform config. 

NB: recommend redirect id 39129 is deleted (`form/contact-the-make-the-call-team`) as otherwise saving/importing config can clash with a redirect that covers that path.


omahm (approved):
LG

PR #708, opened 2021-05-24
Add patch and config for preferred countries

johangant:
* For contacts addresses
* GP practices restricted to UK only so not needed on that entity type.


neilblair (approved):
LG

PR #717, opened 2021-06-01
footer text (field_additional_info)

neilblair:
Add footer text (field_additional_info) to publications content types.
Set footer text in full display to use ancestral value format on both themes and publications.


DuttonMa (approved):
All looks good.
Presumably there is another PR for the theme change ?

neilblair:
> All looks good.
> Presumably there is another PR for the theme change ?

https://github.com/dof-dss/nicsdru_nidirect_theme/pull/291

PR #730, opened 2021-06-11
Catch-all routing redirect config

johangant:
Should help address the rare instances where Drupal can generate a base URL of /index.php/pathname.


omahm (approved):
🤞

PR #733, opened 2021-06-14
Add routing rule for webmanifest files in custom themes directory

johangant:


johangant:
image

Now showing HTTP 200 response code for the webmanifest file

johangant:
Charging ahead a bit with this so it's ready for general QA tomorrow.

PR #734, opened 2021-06-15
D8NID-1288 news RSS feed

johangant:
* Omitted flag values as there was no query filtering on them, only extra joins.
* Preprocess code added to theme to overcome some shortcomings from Core.

Relates to https://github.com/dof-dss/nicsdru_nidirect_theme/pull/300


omahm (approved):
LG

PR #740, opened 2021-06-24
Webforms fixes

neilblair:
Many webform fixes - essential for UAT of webforms to be completed this week ...

Includes GDPR related fixes
 - webforms collecting personal information should not store it in the db - so we set results_disabled: true
 - quiz style webforms may need to save results to function - and although they don't collect personal info, they do collect IP address which is considered personal info under GDPR - so we set form_disable_remote_addr: true

Also closes the contact form that comes with the webform module - it can be exploited as an open mail relay - upgraded version fixes - https://www.drupal.org/project/webform/releases/6.0.2


config/sync/webform.webform.contact.yml
```diff
@@ -1,6 +1,6 @@
 uuid: cc3f4d2f-b0ab-4121-a21d-9cc8a17aa139
 langcode: en
-status: open
+status: closed
```
johangant, line 3: Intended to be closed?

config/sync/webform.webform.contact.yml
```diff
@@ -208,7 +228,7 @@ handlers:
     label: 'Email confirmation'
     notes: ''
     handler_id: email_confirmation
-    status: true
+    status: false
```
johangant, line 231: Do these need to be disabled if the webform is closed as a whole, or the environment they're running on isn't connected to an operating MTA service? Could be that they're to be all switched on a little later down the line but wanted to check

config/sync/webform.webform.debt_management_enquiry.yml
```diff
@@ -196,10 +196,10 @@ settings:
   wizard_confirmation: true
   wizard_confirmation_label: ''
   wizard_track: ''
-  preview: 0
+  preview: 2
   preview_label: ''
   preview_title: ''
-  preview_message: ''
+  preview_message: '<span style="-webkit-text-stroke-width:0px; color:#000000; display:inline !important; float:none; font-family:Lato,&quot;Open Sans&quot;,sans-serif; font-size:21px; font-style:normal; font-variant-caps:normal; font-variant-ligatures:normal; font-weight:400; letter-spacing:normal; orphans:2; text-align:start; text-decoration-color:initial; text-decoration-style:initial; text-decoration-thickness:initial; text-indent:0px; text-transform:none; white-space:normal; widows:2; word-spacing:0px">Before sending information, check the information is accurate.&nbsp;</span>'
```
johangant, line 202: Might it be easier/more manageable to use a CSS class set in the wider nidirect theme over inline styles in the webform config?

config/sync/webform.webform.contact.yml
neilblair: Yes - it's a security vulnerability - as per PR description

config/sync/webform.webform.contact.yml
neilblair: I disabled this initially and then decided to close the contact form as a whole.  (The form is provided with the Webform module - but we do not use it.  It is in fact a sec vuln because it can act as an open relay)

config/sync/webform.webform.debt_management_enquiry.yml
neilblair: Dodgy cut and paste into the editor field ... will fix

PR #776, opened 2021-07-08
D8 nid 1350

omahm:


web/healthcheck.php
```diff
@@ -0,0 +1,81 @@
...
+      $errors[] = 'Could not delete newly created file in the files directory.';
+    }
+  } else {
+    $errors[] = 'Could not create a file in the files directory.';
+  }
+}
+
+// Create and return a new Response.
+$response = new Response();
+$response->setExpires(new DateTime('01/15/2001'));
```
johangant, line 62: Is this date significant or need to be a specific value?

web/healthcheck.php
```diff
@@ -0,0 +1,81 @@
...
+$response->setLastModified(new DateTime());
+
+// Manually set headers for cache control.
+$response->headers->set('Cache-Control', ['no-store', 'must-revalidate', 'stale-while-revalidate=0', 'stale-if-error=0']);
+$response->headers->set('Surrogate-Control', 'max-age=30');
```
johangant, line 67: Does this mean the file is only executed by the backend every 30 seconds even if requested more than that?

johangant (approved):
Looks good. A few small questions rather than issues but otherwise GTG

web/healthcheck.php
omahm: Just needs to be in the past and this will answer your first query: https://en.wikipedia.org/wiki/Drupal

web/healthcheck.php
omahm: That's what is in the D7 version. @neilblair are your happy with this valiue?

neilblair: The surrogate-control allows fastly to cache the healthcheck script for 30 seconds.  So that provides some level of DoS protection.  I completely forgot this was in place on the D7 version and perhaps resolves our concerns about the script being hit by bots?

https://developer.fastly.com/reference/http-headers/Surrogate-Control/

PR #777, opened 2021-07-08
D8 nid 1363

DuttonMa:
Install dblog module


config/sync/ultimate_cron.job.update_cron.yml
```diff
@@ -0,0 +1,17 @@
+uuid: bea581c7-b334-423c-b122-aa68c873c937
+langcode: en
+status: true
+dependencies:
+  module:
+    - update
+title: 'Update indexes'
```
johangant, line 7: Is this required for dblog to operate?

config/sync/dblog.settings.yml
```diff
@@ -0,0 +1,3 @@
+row_limit: 1000
```
johangant, line 1: I'd suggest a slightly higher amount... 10k? CSP reports would fill this rather quickly in the short term. 10k rows shouldn't take up much disk space on the db service.

config/sync/ultimate_cron.job.update_cron.yml
DuttonMa: Good question, I'll try deleting it

config/sync/dblog.settings.yml
DuttonMa: Yes, another good spot !

johangant (approved):
LGTM

PR #780, opened 2021-07-14
D8 nid 1363

DuttonMa:
Install nidirect_dblog module and override dblog_cron so that certain log entries are not deleted.

See associated PR https://github.com/dof-dss/nidirect-site-modules/pull/422

### DIFF
--- a/config/sync/dblog.settings.yml
+++ b/config/sync/dblog.settings.yml
@@ -1,3 +1,3 @@
-row_limit: 1000
+row_limit: 10000
 _core:
   default_config_hash: e883aGsrt1wFrsydlYU584PZONCSfRy0DtkZ9KzHb58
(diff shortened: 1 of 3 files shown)


omahm (approved):

config/sync/dblog.settings.yml
```diff
@@ -1,3 +1,3 @@
-row_limit: 10000
+row_limit: 1000
```
johangant, line 1: Is the reduction here intentional? I reckon we can spare an extra 9000 db rows with all the disk space we have available :)

DuttonMa: Well, it was intentional, my thinking was that we were now being more selective about what we kept for the long term, so we didn't need to use such a 'shotgun approach' !
But maybe you're right, I'll change it back so that we can have our cake and eat it !

PR #790, opened 2021-08-05
CI refinements and refactoring

johangant:
Annotations in comments for better clarity! Bit of a YML-grenade, otherwise.

NB: once merged, an overnight test will be needed to confirm the workflows execute as they should. Any GH branch protection rules might also need a tweak in case job names have changed a little.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -286,9 +286,6 @@ jobs:
             # Push to fixed, non-integrating build branch. GitHub webhook integration will propagate this
             # to platform.sh for later steps to use.
             git push -f origin $EDGE_BUILD_BRANCH
-
-            # Pause for webhook to propagate the change from GH to PSH.
-            sleep 10s
       - run:
           name: Activate the environment if not found.
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,6 +43,15 @@ commands:
           name: Fetch dependencies with composer
           command: |
             composer install --no-interaction --optimize-autoloader
+  composer_tasks__edge_packages:
+    description: "Switch dof-dss packages to HEAD on development branch"
+    steps:
+      - run:
+          name: Switch dof-dss packages to HEAD on develelopment branch
+          command: |
+            composer require dof-dss/nidirect-site-modules:dev-development dof-dss/nidirect-migrations:dev-development \
+              dof-dss/nidirect-d8-test-install-profile:dev-development dof-dss/nicsdru_origins_theme:dev-development \
+                dof-dss/nicsdru_origins_modules:dev-development dof-dss/nicsdru_nidirect_theme:dev-development
   install_php_os_extensions:
     description: "Install a standard group of extensions and packages"
     steps:
@@ -52,36 +61,32 @@ commands:
             sudo cp /home/circleci/project/.circleci/docker-php-circleci.ini /usr/local/etc/php/conf.d/
             sudo apt-get update
             sudo apt-get install -y libpng-dev
-            sudo docker-php-ext-install gd
-  install_platformsh_tools:
-    description: "Install packages and tools for interacting with platform.sh API"
+            sudo docker-php-ext-install gd pcntl posix
+  hosts_keyscan:
+    description: "Keyscan for hosts that require SSH access"
     steps:
       - run:
-          name: Add extra OS and PHP extensions/config
-          command: |
-            sudo docker-php-ext-install pcntl posix
-      - run:
-          name: Keyscan Platform.sh region hostnames + GitHub for password-less access
+          name: Keyscan for hosts that require SSH access
           command: |
             mkdir -p ~/.ssh
-            ssh-keyscan -H ssh.$PLATFORM_REGION >> ~/.ssh/known_hosts
             ssh-keyscan -H github.com >> ~/.ssh/known_hosts
-      - run:
-          name: Install the Platform.sh CLI tool
-          command: |
-            curl -sS https://platform.sh/cli/installer | php
-  github_hosts_workaround:
-    description: "Adds github.com to known hosts in container; for working locally."
+            ssh-keyscan -H ssh.$PLATFORM_REGION >> ~/.ssh/known_hosts
+  install_psh_cli:
+    description: "Install the Platform.sh CLI tool"
     steps:
       - run:
-          name: Keyscan GitHub hostname
-          command: mkdir -p ~/.ssh && ssh-keyscan -H github.com >> ~/.ssh/known_hosts
+          name: Install the Platform.sh CLI tool
+          command: curl -sS https://platform.sh/cli/installer | php
+      - run:
+          name: Add platform cli tool to $PATH
+          command: echo 'export PATH="$HOME/"'.platformsh/bin':"$PATH"' >> $BASH_ENV
 jobs:
   # Tests the integrity of the build, stores the results in a workspace for re-use in later jobs.
   build:
     docker:
       - image: circleci/php:7.4.15-apache-browsers
     steps:
+      - hosts_keyscan
       - checkout_code
       - install_php_os_extensions
       - composer_tasks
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -43,6 +43,15 @@ commands:
           name: Fetch dependencies with composer
           command: |
             composer install --no-interaction --optimize-autoloader
+  composer_tasks__edge_packages:
```
johangant, line 46: For context: this was previously done like this:

* CI workflow scheduled to run M-F on the `development` branch.
* A job would then switch over the checked out branch (`development`), get composer to update to latest commits on repo branches, then commit + push back to GH.
* Really all we needed to do here was: in the workflow where we build the edge environment, set it to listen/pull on the edge branch (`D8NID-edge`), then fetch the latest commits, then push back. It's a bit simpler that way and avoids some mental gymnastics to understand what code is in operation at what stage in the pipeline.

.circleci/config.yml
```diff
@@ -52,89 +61,78 @@ commands:
...
-          command: |
-            sudo docker-php-ext-install pcntl posix
-      - run:
-          name: Keyscan Platform.sh region hostnames + GitHub for password-less access
+          name: Keyscan for hosts that require SSH access
           command: |
             mkdir -p ~/.ssh
-            ssh-keyscan -H ssh.$PLATFORM_REGION >> ~/.ssh/known_hosts
             ssh-keyscan -H github.com >> ~/.ssh/known_hosts
+            ssh-keyscan -H ssh.$PLATFORM_REGION >> ~/.ssh/known_hosts
```
johangant, line 73: I'm not sure we need to be coy about what region we host on... it certainly wouldn't be too hard to make an educated guess or two to find out. Either way, this is left as a variable value in case something changes in future.

.circleci/config.yml
```diff
@@ -52,89 +61,78 @@ commands:
...
+  install_psh_cli:
+    description: "Install the Platform.sh CLI tool"
```
johangant, line 75: Made this a command as it's good to have a fixed way of doing this. Used to be added to individual jobs as run steps within them, this is a little bit more DRY

.circleci/config.yml
```diff
@@ -52,89 +61,78 @@ commands:
...
-            curl -sS https://platform.sh/cli/installer | php
-  github_hosts_workaround:
-    description: "Adds github.com to known hosts in container; for working locally."
-    steps:
+          command: curl -sS https://platform.sh/cli/installer | php
       - run:
-          name: Keyscan GitHub hostname
-          command: mkdir -p ~/.ssh && ssh-keyscan -H github.com >> ~/.ssh/known_hosts
+          name: Add platform cli tool to $PATH
+          command: echo 'export PATH="$HOME/"'.platformsh/bin':"$PATH"' >> $BASH_ENV
```
johangant, line 82: Significantly simplifies using the platform CLI tool. Took me a while to understand how to get this stored for other things to use; a quirk of the CI service I guess.

.circleci/config.yml
```diff
@@ -52,89 +61,78 @@ commands:
...
+    environment:
+      PROJECT_ROOT: "/home/circleci/project"
+      PLATFORM_REGION: "uk-1.platform.sh"
     steps:
       - attach_workspace:
           at: ./
       - run: sudo cp /home/circleci/project/.circleci/docker-php-circleci.ini /usr/local/etc/php/conf.d/
       - run:
           name: PHPCS analysis
-          command: /home/circleci/project/phpcs.sh /home/circleci/project "/home/circleci/project/web/modules/origins /home/circleci/project/web/modules/custom /home/circleci/project/web/modules/migrate /home/circleci/project/web/themes/custom"
```
johangant, line 124: Used a bash variable to build up this list. It's much clearer to read and easier to maintain.

.circleci/config.yml
```diff
@@ -265,11 +263,13 @@ jobs:
       EMAIL: "circleci@localhost"
       GIT_COMMITTER_NAME: "Circle CI"
       GIT_AUTHOR_NAME: "Circle CI"
+      EDGE_BUILD_BRANCH: "D8NID-edge"
```
johangant, line 266: Opted for clarity over security through obscurity here. It's no stretch of the imagination to see which code branch provides the edge build in a public repo, so it's easier setting that variable value here rather than injecting it through Circle CI's env vars options which, annoyingly, you can't see the value of once they're set.

.circleci/config.yml
```diff
@@ -306,60 +306,54 @@ jobs:
...
-      - run:
-          name: Checkout edge branch
-          command: |
-            cd ~/project
-            git checkout -b $EDGE_BUILD_BRANCH
       - install_php_os_extensions
-      - install_platformsh_tools
+      - install_psh_cli
       - run:
-          name: Trigger a data sync from production environment to nightly edge build.
```
johangant, line 318: Dropped the nightly references as the cron syntax indicates frequency. Sticking to just 'edge' as a way of designating the edge build/environment.

.circleci/config.yml
```diff
@@ -306,60 +306,54 @@ jobs:
...
           name: Refresh configuration as our db will contain active prod config after sync operation
           command: |
-            /home/circleci/.platformsh/bin/platform environment:drush cim -y -p $PLATFORM_PROJECT_ID -e $EDGE_BUILD_BRANCH
-      - run:
-          name: Run structure sync import all
-          command: |
-            /home/circleci/.platformsh/bin/platform environment:drush "ia --choice=safe" -y -p $PLATFORM_PROJECT_ID -e $EDGE_BUILD_BRANCH
+            platform environment:drush cim -y -p $PLATFORM_PROJECT_ID -e $EDGE_BUILD_BRANCH
       - run:
           name: Rebuild Taxonomy Entity Index
```
johangant, line 326: Can't for the life of me remember what this is for, but it's working and it's been there for a while so.... if it works, don't fix it etc

.circleci/config.yml
```diff
@@ -369,12 +363,9 @@ workflows:
       - deprecated_code:
           requires:
             - build
-      - unit_kernel_tests:
-          requires:
-            - build
 
-  # A nightly build of the project, using all dof-dss packages at HEAD from development branch.
-  nightly-edge-build:
+  # A daily build of the project, using all dof-dss packages at HEAD from development branch.
```
johangant, line 367: Left daily in the comment, which matches the cron syntax. No reason why it should be nightly... it might execute in sunlit hours at some point in the future... who knows!

.circleci/config.yml
```diff
@@ -384,37 +375,22 @@ workflows:
...
                 - development
     jobs:
-      - build
-      - coding_standards:
-          requires:
-            - build
-      - deprecated_code:
-          requires:
-            - build
-      - unit_kernel_tests:
```
johangant, line 394: Banished this long running, much despised (by me!) step, as our build integrity is going to be verified nightly in our edge build process.

There's a chance something might break and we won't know until we try to build it... but the edge build is volatile anyway so if we need to sit, fix + push it up to rebuild it, then so be it. By shifting this step into the edge build process, we speed up routine code commits and PRs.

.circleci/config.yml
```diff
@@ -384,37 +375,22 @@ workflows:
...
-          requires:
-            - build
+      - edge_build
 
-  # A separate scheduled workflow to sync the data after the nightly build completes.
+  # A separate scheduled workflow to sync the data after the edge build completes.
   # This is separate to avoid using build time/compute units with arbitrary sleep commands;
   # platform.sh deploy hooks can take a few minutes to execute after commits are relayed to it.
-  nightly-content-sync:
+  edge-build-post-build-tasks:
```
johangant, line 383: Renamed because it does a bit more than just sync content. It's basically a workflow to happen after the edge build has run. Would be nice if we could make it dependent on the success of another workflow but Circle CI doesn't stretch to that level... just dependencies between jobs in the same workflow.

.circleci/config.yml
```diff
@@ -52,207 +61,125 @@ commands:
...
           at: ./
       - run: sudo cp /home/circleci/project/.circleci/docker-php-circleci.ini /usr/local/etc/php/conf.d/
       - run:
           name: Deprecated code check
           command: |
-            cd /home/circleci/project
-            vendor/bin/drupal-check /home/circleci/project/web/modules/custom /home/circleci/project/web/modules/origins /home/circleci/project/web/modules/migrate /home/circleci/project/web/themes/custom -e "*/tests/src/*"
-
-  # Run any unit tests and any kernel tests against a vanilla D8 site + our site config imported over it (no predefined content).
-  unit_kernel_tests:
```
johangant, line 140: Removed (for now) as we can run these as part of our edge build, or in a dedicated nightly workflow

omahm (approved):
Looks good. Nice commenting

PR #793, opened 2021-08-12
A11y fixes

neilblair:
- Set search label and placeholder so that they are consistent - note spacing in label is deliberate to ensure correct pronunciation by screen readers
- Change extra_comments_for_pooh_bear into a textfield instead of textarea because regex pattern attributes added to textareas are not valid HTML


omahm (approved):
LG

PR #795, opened 2021-08-13
Remove regex pattern validation on textareas in various webforms

neilblair:
The pattern attribute on textareas is usually used to prevent submission of html - and browsers seem to support it ok - but it is not allowed in the html spec ...

### DIFF
--- a/config/sync/webform.webform.contact_the_make_the_call_team.yml
+++ b/config/sync/webform.webform.contact_the_make_the_call_team.yml
@@ -159,8 +159,6 @@ elements: |
       '#description_display': before
       '#required': true
       '#required_error': 'You must explain why you cannot make contact by phone'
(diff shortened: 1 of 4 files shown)


omahm:
This undoes the requirements from [https://digitaldevelopment.atlassian.net/browse/D8NID-1168](https://digitaldevelopment.atlassian.net/browse/D8NID-1168). Are we ok with that?

neilblair:
> This will undo the requirements from https://digitaldevelopment.atlassian.net/browse/D8NID-1168. Are we ok with that?

Not entirely sure to be honest.  The original intention was to discourage spam but the regex employed only prevents HTML - so I don't think it is really doing anything to discourage spam?  Might be a good idea to have a chat and think about alternatives.

PR #796, opened 2021-08-18
Set autocomplete on input fields collecting information about the user

neilblair:
This is to meet https://www.w3.org/TR/WCAG21/#identify-input-purpose - see https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html


johangant (approved):
LGTM

johangant:
Think if you pull the latest from `development` it will correct the build step in the CI workflow.

PR #800, opened 2021-08-20
Metatags defaults update

johangant:
For:

* Front page
* Global defaults (eg: views pages)
* GP practice nodes
* Landing page nodes
* Taxonomy term pages


omahm (approved):
Nice one

PR #808, opened 2021-09-17
Update core to 8.9.19 + contrib

johangant:
Fixes SA-CORE-2021-006 to 010.

Contrib TBC


johangant:
Omitted some of the more involved contrib updates that might otherwise benefit from closer review/QA from the QAT team.

PR #860, opened 2021-12-06
D8NID-1453

DuttonMa:
Ensure that edge config still runs after sync failure


omahm (approved):
LG

johangant (approved):
Looks good, let's see how it works in production :)

PR #864, opened 2021-12-07
Fix issue where home page "Popular information" is listing popular services

neilblair:
Updates popular content views so that:
- "information site wide" shows articles, contacts, health conditions, landing pages and news (application removed)
- "services site wide" shows application pages only (cold weather payment removed - the cold weather payment checker is an application page)
- "by top-level term" and "by term" show articles, health conditions and landing pages only

### DIFF
--- a/config/sync/views.view.popular_content.yml
+++ b/config/sync/views.view.popular_content.yml
@@ -5,10 +5,10 @@ dependencies:
   config:
     - node.type.application
     - node.type.article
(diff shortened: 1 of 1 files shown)


omahm (approved):
🤦‍♂️ ..LG

neilblair:
> 🤦‍♂️ ..LG

No-one noticed this 😂  I happened to notice it when looking at something else!

PR #874, opened 2022-01-20
D8 nid 1464

neilblair:
Fixes an issue with the Report Legal Aid Fraud webform where the generated email contains duplicate headings. Instead of using [webform_submission:values] token to output everything, we selectively output specific elements to avoid outputting wizard elements whose title is the same as a fieldset that wraps elements within.


omahm (approved):
LG

PR #900, opened 2022-04-05
Modified approach for sending logs to logz 

neilblair:
Previous approach of sending logs to logz using Filebeat has been abandoned.  Too many gotchas and issues with that approach
- starting Filebeat as background process via cron causes unending crons that block deployments
- truncation of log files by Platform causes Filebeat to lose track of log file entries and leads to duplicate entries
- running Filebeat "once" (so it does not constantly run) to upload batched log entries from copies of log files (which are not truncated) is not viable because of unfixed bugs with this mode of operation in Filebeat that cause Filebeat to continue running (and again unending crons and blocked deployments 👎🏻 )

In this PR:
- Filebeat stuff completely removed
- Instead, a bash script is used to batch upload new log entries (written to /app/log mounted directory) via cURL
- Installed filelog module to capture drupal watchdog entries to log file (also written to /app/log mounted directory)
- Syslog module uninstalled (not compatible with Platform environment read only file system)
- New entries to /var/log/access.log and /app/log/drupal.log are captured and uploaded to logz.io every 5 minutes.  Other logs can be added later if needed.


johangant (approved):
Looks great, nicely written up PR

PR #907, opened 2022-05-12
Add missing metatags field to webform content type

neilblair:
Need to be able to edit meta tags on webform nodes - e.g. to hide some forms from search engine indexing


johangant (approved):
LGTM

PR #915, opened 2022-06-07
D8 nid 1515

omahm:


.platform/solr_config/solrcore.properties.old
```diff
@@ -0,0 +1,13 @@
+solr.replication.master=false
```
johangant, line 1: Not sure we need this .old file? Could well be autogenerated as a fallback, if needed.

omahm:
> All good, but I think we're missing the equivalent of this file: https://github.com/dof-dss/nicsdru_dept/blob/development/.platform/solr_config/solrcore.properties
> 
> You can probably copy/paste that file as the config values look right. For full(er) details, https://docs.platform.sh/guides/drupal9/solr.html#6-export-and-modify-configuration shows the key things to adjust.

That's weird it's showing as empty when it has content in the branch:
https://github.com/dof-dss/nidirect-drupal/blob/D8NID-1515/.platform/solr_config/solrcore.properties

PR #928, opened 2022-06-23
"Pages with no theme" view config

neilblair:
For content types where a theme/subtheme is a mandatory field (articles, publications, application pages, etc), if editors delete a theme then pages tagged with the theme do not display correctly.  This PR creates a content view that lists pages with no theme so that Editors can edit the affected pages and add a theme.

### DIFF
--- a/config/sync/views.view.pages_with_no_theme.yml
+++ b/config/sync/views.view.pages_with_no_theme.yml
@@ -1,0 +1,854 @@
+uuid: 00d312e9-ba46-487b-96ca-eccb43550a2e
+langcode: en
+status: true
+dependencies:
+  config:
+    - core.entity_view_mode.node.teaser
+    - node.type.application
+    - node.type.article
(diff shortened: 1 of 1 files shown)


config/sync/views.view.pages_with_no_theme.yml
```diff
@@ -0,0 +1,854 @@
...
+      contexts:
+        - 'languages:language_content'
+        - 'languages:language_interface'
+        - url
+        - url.query_args
+        - user
+        - 'user.node_grants:view'
+        - user.permissions
+      tags: {  }
+  page_1:
```
johangant, line 824: Where possible, it's best to set these display ids to something more descriptive. Helps immensely with debugging and general readability

johangant (changes requested):
Minor correction requested around display id to be changed from it's default. Otherwise, looks good and helpful!

PR #952, opened 2022-08-31
D8NID-1540

neilblair:
- enable nidirect_services module (https://github.com/dof-dss/nidirect-site-modules/pull/480)
- changes to services catalog view that generates services-catalogue.json to enable custom serializer that converts relative urls in field content to absolute urls


omahm (approved):
LG

PR #982, opened 2022-10-25
D8NID-1558 ~ Update Sitewide alerts module. Apply Redis patch to fix …

omahm:
…bug with sitewide alerts


johangant (approved):
Nice spot! LGTM

PR #995, opened 2022-11-23
D8NID-1563 core 9.4 config export

johangant:
Adds missing config into core.extensions.yml

Rest of config has been sorted a little per file but remains the same.


neilblair (approved):
Thanks Johan 👍🏻

PR #997, opened 2022-12-07
Add json export display for driving instructors view

neilblair:


omahm (approved):

config/sync/views.view.driving_instructors.yml
```diff
@@ -427,3 +440,891 @@ display:
         - 'user.node_grants:view'
       tags:
         - 'config:search_api.index.contacts'
+  rest_export_1:
```
johangant, line 443: Would be ideal to have a more descriptive display id. The defaults views produces can be ambiguous and easily missed in the config UI

johangant (approved):
Looks good, just the nit-pick about the display ids from me

PR #1071, opened 2023-10-30
D8NID-1606 ~ Update template path for classy theme

omahm:


johangant (approved):
LGTM. Same thing needed on DEPT not that long ago

PR #1073, opened 2023-10-30
D8NID-1602 workaround for badly handled oembed iframe assets

johangant:
https://www.drupal.org/project/drupal/issues/3350299 sums up the issue nicely. The patch/issue is only for olivero theme rather than Claro but am sure this will be ported in due course.

For now, this fits in well with our CKEditor overrides and tweaks.


johangant:
Closing with a big NOPE while I get this mess sorted out

PR #1098, opened 2023-11-09
D8 nid 1623

omahm:
Shifting from Listbuilder to the bundled (and maintained) View with Redirect. ListBuilder patch was missing the accessCheck() call required in 9.2+ and I don't want to rely on this patch being updated for any future core changes.


dof-dss-eddwebdev (approved):

johangant (approved):
Thanks, this looks good. Interesting to see the patch needed for ultimate_cron module too

PR #1112, opened 2023-11-21
D8 nid 1619

omahm:


.platform/solr_config/solrcore.properties
```diff
@@ -1,16 +1,14 @@
 solr.replication.master=false
 solr.replication.slave=false
 solr.replication.pollInterval=00:00:60
-solr.replication.masterUrl=http://solr.internal:8080/solr
+solr.replication.masterUrl=http://localhost:8983/solr
```
johangant, line 4: Is this right? Platform.sh would usually use port 8080 (hostname might vary with the service name change?) but this one looks like a Lando service which typically runs on port 8983

johangant (changes requested):
LGTM, just the one possible red flag around the service hostname/port defined in the .platform config directory

.platform/solr_config/solrcore.properties
omahm: This is identical to the comment I left on Martin's PR and this morning on standup he said that 8983 oddly enough works.

johangant: Okay, great. If it works it works :)

johangant (approved):

PR #1153, opened 2024-02-02
Prison visits virtual from json

neilblair:
Alter prison visit booking handler to retrieve all visit slots, face-to-face slots and virtual visit slots, from data posted from PRISM.  (Previously face-to-face slots came from PRISM data and virtual slots were notionally created from config)

Add a submission serial number to emailed form submission to help identify when a submission has not been received.

### DIFF
--- a/web/modules/custom/nidirect_webforms/src/Plugin/WebformHandler/PrisonVisitBookingHandler.php
+++ b/web/modules/custom/nidirect_webforms/src/Plugin/WebformHandler/PrisonVisitBookingHandler.php
@@ -89,6 +89,18 @@ class PrisonVisitBookingHandler extends WebformHandlerBase {
 
   /**
    * {@inheritdoc}
+   *
+   * If webform submissions are optionally enabled, we can prevent or
+   * alter the submission before insertion into the database to
+   * obfuscate sensitive data or remove it altogether.
+   */
+  public function postSave(WebformSubmissionInterface $webform_submission, $update = TRUE) {
+    // For now, just delete the submission entirely!
+    $webform_submission->delete();
+  }
+
+  /**
+   * {@inheritdoc}
    * @throws \Exception
    */
   public function alterForm(array &$form, FormStateInterface $form_state, WebformSubmissionInterface $webform_submission) {
@@ -378,13 +390,6 @@ class PrisonVisitBookingHandler extends WebformHandlerBase {
         }
       }
     }
-  }
-
-  /**
-   * {@inheritdoc}
-   */
-  public function postSave(WebformSubmissionInterface $webform_submission, $update = TRUE) {
-    $webform_submission->delete();
   }
 
   /**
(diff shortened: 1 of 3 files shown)


web/modules/custom/nidirect_webforms/src/Plugin/WebformHandler/PrisonVisitBookingHandler.php
```diff
@@ -800,28 +763,54 @@ private function getAvailableSlots() {
...
-        $updated_slots[] = $slot_date_adjusted;
+    }
+    else {
+
+      // No data in cache, so try file instead. Every time the
+      // external service posts data to prison visits api controller,
+      // data is stored in cache and written to file.
+
+      $file_uri = 'private://nidirect_webforms/prison_visit_slots_data.json';
+      $file_contents = @file_get_contents($file_uri);
```
omahm, line 802: Using @ is not recommended. I'd use file_exists to check the file and then read the contents.

web/modules/custom/nidirect_webforms/src/Plugin/WebformHandler/PrisonVisitBookingHandler.php
```diff
@@ -800,28 +763,54 @@ private function getAvailableSlots() {
...
+
+      if (!empty($file_contents)) {
+        $data = json_decode($file_contents, TRUE);
```
omahm, line 805: PHP 8 introduces the json_validate method that can be used to check if the file contents are valid JSON and and errors can be retrieved with json_last_error()
https://www.php.net/manual/en/function.json-validate.php

omahm (changes requested):
Looks good, just a couple of defensive coding suggestions.

web/modules/custom/nidirect_webforms/src/Plugin/WebformHandler/PrisonVisitBookingHandler.php
neilblair: The JSON is validated on the way in by the [PrisonVisitBookingJsonApiController.php](https://github.com/dof-dss/nidirect-drupal/blob/prison-visits-virtual-from-json/web/modules/custom/nidirect_webforms/src/Controller/PrisonVisitBookingJsonApiController.php) route controller around line 188.  

So the data is only saved to cache and to the file if it is valid.  Is that enough or should it be validated when retrieved as well?

omahm: Looks good 👍

neilblair: I suppose there might be a possibility the file becomes corrupted ...?

omahm (approved):

PR #1191, opened 2024-04-25
Shift front-end tooling out of appserver container into own

johangant:
Makes it easier to isolate these tools in a versioned container rather than bundling them into the main appserver container. Also reduces the reliance on host installed tooling making this more re-usable across platforms and the wider team.

See https://github.com/dof-dss/nicsdru_unity/pull/1947#issue-2263719105 for evaluation details

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -24,16 +24,8 @@ compose:
 services:
   appserver:
     scanner: false
-    environment:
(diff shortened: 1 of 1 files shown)


johangant:
Closing due to DDEV support providing npm/node tooling from within the web container. Ref: https://github.com/dof-dss/nicsdru_dept/pull/891/files

PR #1196, opened 2024-05-09
Update prison booking webform

neilblair:
Update PrisonVisitBookingHandler.php

Set a flag for JS to reset timeslot preferences and also have to clear form state and element default values / data serverside as well.

Update prisonbooking.js

- Reset timeslots when settings.prisonVisitBooking.resetTimeslots === true
- Some other bugfixes and refactoring


omahm (approved):
LG

PR #1201, opened 2024-05-22
Update telephone_plus module and tweak site specific tel formatter

johangant:
Some slight differences between the way NIDirect's version of telephone_plus' formatter handles extensions. Extending it a bit to check for the unwanted extra phone extensions text seems to help


neilblair (approved):
Approving this downgrade of telephone_plus to resolve D8NID-1667

PR #1209, opened 2024-05-29
D8NID-1660 split hospital emergency department waiting times into two separate tables

neilblair:
Emergency department waiting times should be shown as two tables
- minor injury units / urgent care services
- emergency departments


johangant (approved):
LGTM :+1:

PR #1213, opened 2024-05-30
Update view display config for contact, gp_practice, driving_instructor

neilblair:
Changed field_telephone to use "NIDirect Telephone Link" formatter as "Telephone Plus" formatter does not exist due to downgrade of telephone_plus module from 2.0@beta to 9.1.


johangant (approved):
:+1:

PR #1227, opened 2024-06-03
D8NID-1664: prison visits booking — better DOB entry fields

neilblair:
Standard html date fields when rendered on smart phones force users to use a calendar widget to enter a date and this can be unwieldy for entering dates of birth.  Change DOB fields to plain text fields with input mask for easier and quicker date entry. Since the fields are standard text fields, they need custom date validation.

Screenshot 2024-08-19 at 10 02 25

Screenshot 2024-08-19 at 10 03 52

Screenshot 2024-08-19 at 10 05 35


omahm (approved):
LGTM 👍

PR #1232, opened 2024-06-10
Security headers update

neilblair:
Add missing `X-Content-Type-Options: nosniff` for static files.

Add/Update `Strict-Transport-Security`.  Oddly this header is present on nidirect despite there being no config for it in routes.yaml ... So perhaps it is being added by Fastly.


johangant (approved):
Looks like a sensible change to me

PR #1268, opened 2024-08-19
D8 nid 1704

omahm:
- Allow the removal of previously set URL's
- Replace link URI fields set to entity reference with textfield
- Automatically add and remove titles for certain link fields.


johangant (approved):
Not sure why the link fields are/were set to entity reference (can we changed them to text type in config?) but I see they're handled here for now

PR #1359, opened 2025-03-26
D8NID-1699 Fix Mysterious Missing Breadcrumb

neilblair:
After many months of disappearing breadcrumb trails, it turns out that an early return in NodeThemesBreadcrumb.php was causing (I think) an empty breadcrumb to be cached with no cache context leading to all nodes then showing an empty breadcrumb after a user views a feature/featured_content_list.


omahm (approved):
LGTM

johangant (approved):

PR #1382, opened 2025-06-25
D8 nid 1774

omahm:
Adding form elements and div attributes to Basic HTML as edits to the CWP page break the form rendering


johangant (approved):
LGTM

PR #1387, opened 2025-07-03
D8NID-1782 Prison Visits: fixes for chosen time slots getting dropped and incorrect preference ranks being shown

neilblair:
Fix issue where timeslot ranks were not displaying correctly and tracking of preferred timeslots was not behaving correctly when user hits previous to change already selected timeslots.
- Now using once() wrapped around all of the behaviour code
- Using click event on time slots instead of change event ensures other webform behaviours that make changes to checkboxes are not triggering unexpected time slot rank changes


omahm (approved):
I'll not pretend to know exactly what is going on but in principle it looks good.

PR #1400, opened 2025-07-21
Make email handlers conditional based on selected values

johangant:
Debt recovery form should send submissions to specific addresses depending on what form values have been selected by the user


johangant:
NB: deploy by manual importing via the UI due to config_ignore being in operation on drush cim

PR #1430, opened 2025-09-05
D8NID-812 Prison visits: additional visitors added to a virtual booking

neilblair:
Fixes issue where if the user makes a face to face booking with additional visitors, and then goes back and changes the booking to a virtual booking, additional visitors were not being removed from the virtual booking.

Also fixed another issue I came across where remembered additional visitor details were being wiped if a user makes a virtual booking immediately after making a face-to-face booking.

### DIFF
--- a/web/modules/custom/nidirect_webforms/src/Plugin/WebformHandler/PrisonVisitBookingHandler.php
+++ b/web/modules/custom/nidirect_webforms/src/Plugin/WebformHandler/PrisonVisitBookingHandler.php
@@ -609,6 +609,7 @@ class PrisonVisitBookingHandler extends WebformHandlerBase {
     $temp_store = $this->tempStoreFactory->get('nidirect_webforms.prison_visit_booking');
 
     // Retrieve visitor data, but not for virtual bookings.
+    $visitor_data = [];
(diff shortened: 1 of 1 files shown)


web/modules/custom/nidirect_webforms/src/Plugin/WebformHandler/PrisonVisitBookingHandler.php
```diff
@@ -265,12 +268,37 @@ public function alterForm(array &$form, FormStateInterface $form_state, WebformS
     $form['#attached']['drupalSettings']['prisonVisitBooking'] = $this->configuration;
     $form['#attached']['drupalSettings']['prisonVisitBooking']['booking_ref'] = $this->bookingReference;
 
+    // Detect if booking reference (visit_order_number) is changed.
+    $last_ref = $form_state->get('last_visitor_order_number');
+    $curr_ref = $form_state->getValue('visitor_order_number');
+
+    // First-time set.
+    if ($last_ref === NULL && !empty($curr_ref)) {
```
johangant, line 276: Safer to check for non-values with `empty()` instead of NULL, which is a far more specific type and (non)-value check

johangant (approved):
Looks good, just a few very small nitpicks to review

PR #1436, opened 2025-09-10
D8NID-1818

neilblair:
Create a service status rest endpoint for prison visit booking form to enable Prism to open or close the prison visit booking webform and set a service unavailable message.


johangant (approved):
LGTM 👍

PR #1455, opened 2025-09-24
Fix linkit inserting aliased instead canonical urls in links

neilblair:
- Remove patch "Show URL alias after autocomplete selection instead of internal route" because it causes the stored link path to be the alias rather than the canonical path
- Update allowed html to add data-* attributes required for the LinkIt URL Converter to work to convert canonical link paths to aliased paths at render time.


johangant (approved):
👍

PR #1534, opened 2026-03-04
PHP 8.4 config and custom code fixes

johangant:


neilblair (approved):
Will need thorough testing on this especially in relation to beta version of webform

PR #1540, opened 2026-04-10
D8 nid 1894 external contact link icons

johangant:


web/modules/custom/nidirect_contacts/nidirect_contacts.module
```diff
@@ -334,3 +338,86 @@ function nidirect_contacts_camel_to_underscore(string $input_string, string $sep
...
+    // Get the node from the row.
+    $node = $variables['row']->_entity ?? NULL;
+    if ($node instanceof NodeInterface === FALSE) {
+      return;
+    }
+
+    $contact_website_url = $node->get('field_contact_website')->uri ?? '';
+
+    if (str_starts_with($contact_website_url, 'http')) {
+      $variables['external_contact_website_url'] = TRUE;
```
neilblair, line 381: Need to have an explicit check that some kind of option or flag "Show contact as external link" has been checked and website URL provided.

neilblair (changes requested):
Hi Johan,

CET will need an option or flag that allows them to choose whether a contact should be displayed as an external link.  I should have made that clear in the ticket.  Sorry about that.

I was thinking that there should be a checkbox or flag "Show contact as external link" and if selected, the contacts A-Z / search suggestions and results entry would link to the website address for the contact.  

The website address is an optional field at present - so possibly it should become mandatory if the "Show contact as external link" is enabled?

And, (pushing my luck here) if "Show contact as external link" is enabled, all other fields are hidden except the website address field?

Maybe need to discuss what is possible here and best way to do it?

Thanks,
Neil

neilblair (approved):
Great, thanks Johan for implementing those extras 👍🏻

PR #1573, opened 2026-06-03
D8NID-1942 missing prisoner notifications due to premature cancellation of pending transactions

neilblair:
Some additional fixes to prevent visitors cancelling payment transactions in unexpected ways.
- Do not cancel and recreate a visitor's pending transaction when they start a new one because the pending transaction could still be in the process of being authorised by Worldpay and authorisation could arrive after the transaction is cancelled.
- Enable worldpay cancel button to trigger the webform handler's cancelPaymentSubmit() method to more safely cancel a transaction locally.


web/modules/custom/nidirect_prisons/src/Plugin/WebformHandler/PrisonerPaymentsWebformHandler.php
```diff
@@ -312,6 +282,11 @@ public function alterForm(array &$form, FormStateInterface $form_state, WebformS
           '#value' => $this->t('Cancel payment'),
           '#submit' => [[static::class, 'cancelPaymentSubmit']],
           '#limit_validation_errors' => [],
+          '#attributes' => [
+            'aria-hidden' => 'true',
+            'class' => ['visually-hidden'],
```
omahm, line 287: People using a screen reader will still be able to access this button.

neilblair: The aria-hidden attribute will [hide the button from screen readers](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden)

neilblair: Might switch to html hidden attribute as does a similar job - be also hides visually, so wouldn't need the visually-hidden class

neilblair: Nope can't use hidden attribute, doesn't seem to work to visually hide the button as it should (could be overridden by other css).

The combination aria-hidden attribute, visually-hidden class and a tabindex of -1 looks to be best route to make the button completely hidden visually and from screen readers.

omahm: > The aria-hidden attribute will [hide the button from screen readers](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden)

Shoulda went to specsavers !

omahm (approved):

